PR #3, opened 2021-04-21
feat: change login signal to only include user parameter

morenol:
**Description:** This changes the provided parameters of the post login signal to include the request.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Author concerns:**

Still not sure if we need this, we should remove the site parameter and the consider if we need to pass the request or instead get the request with crum.get_current_request in the receivers of the signal.

### DIFF
--- a/openedx_events/signals/auth/v1/__init__.py
+++ b/openedx_events/signals/auth/v1/__init__.py
@@ -8,4 +8,4 @@ from django.dispatch import Signal
 REGISTER_USER = Signal(providing_args=["user", "registration"])
 
 # Signal that fires when a user logins in the platform
-LOGIN_USER = Signal(providing_args=["user", "site"])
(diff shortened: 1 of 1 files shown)


openedx_hooks/signals/auth/v1/__init__.py
```diff
@@ -8,4 +8,4 @@
 REGISTER_USER = Signal(providing_args=["user", "registration"])
 
 # Signal that fires when a user logins in the platform
-LOGIN_USER = Signal(providing_args=["user", "site"])
+LOGIN_USER = Signal(providing_args=["user", "request"])
```
morenol, line 11: Perhaps we only need the "user" parameter and get the request with crum.get_current_request() in the receiver if it is needed.

PR #7, opened 2021-06-08
feat: add necessary tooling for Open edX events 

mariajgrimaldi:
**Description:** 
This PR adds the tooling needed to create and trigger events in Open edX platform. This PR includes:

- OpenEdxPublicSignal class: used to define events with the relevant metadata and methods to trigger them on the Open edX platform. 
- Custom Exceptions: raised by the tooling when there's an error sending or instantiating events

The overall implementation was discussed by members of the community in this PR: #4 

**JIRA:** 
https://edunext.atlassian.net/browse/PS2021-629?atlOrigin=eyJpIjoiODIwNDA5M2MxZGRlNGQ2NjhjNjJhZWIwNWZiNWNlYjUiLCJwIjoiaiJ9

**Testing instructions:**

Option 1: using PR #18 

With the event `STUDENT_REGISTRATION_COMPLETED`:

In your Open edX devstack
1. Go to lms shell `make lms-shell`
2. Install openedx-events specifying the events definition branch:
 `pip install git+https://github.com/eduNEXT/openedx-events.git@MJG/events_definition#egg=events_definition` 
3. Change the Open edX platform branch to [eduNEXT:MJG/1st_batch_openedx_events](https://github.com/edx/edx-platform/pull/28266)
4. Connect your custom receiver to the signal as follows:

```
from openedx_events.learning.signals import STUDENT_REGISTRATION_COMPLETED

@receiver(STUDENT_REGISTRATION_COMPLETED)
def callback(**kwargs):
....
```
We'll add soon an example using our plugin openedx-basic-hooks!
 
**Reviewers:**
- [ ] tag reviewer 
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

### DIFF
--- a/openedx_events/data.py
+++ b/openedx_events/data.py
@@ -1,0 +1,62 @@
+"""
+Data attributes for events within the architecture subdomain `learning`.
+
+These attributes follow the form of attr objects specified in OEP-49 data
+pattern.
+"""
+import socket
+from datetime import datetime
+from uuid import UUID, uuid1
+
+import attr
+from django.conf import settings
+
+import openedx_events
+
+
+@attr.s(frozen=True)
+class EventsMetadata:
+    """
+    Attributes defined for Open edX Events metadata object.
+
+    The attributes defined in this class are a subset of the
+    OEP-41: Asynchronous Server Event Message Format.
+
+    Arguments:
+        id (UUID): event identifier.
+        event_type (str): name of the event.
+        minorversion (int): version of the event type.
+        source (str): logical source of an event.
+        sourcehost (str): physical source of the event.
+        time (datetime): timestamp when the event was sent.
+        sourcelib (str): Open edX Events library version.
+    """
+
+    id = attr.ib(type=UUID, init=False)
+    event_type = attr.ib(type=str)
+    minorversion = attr.ib(type=int, converter=attr.converters.default_if_none(0))
+    source = attr.ib(type=str, init=False)
+    sourcehost = attr.ib(type=str, init=False)
+    time = attr.ib(type=datetime, init=False)
+    sourcelib = attr.ib(type=tuple, init=False)
+
+    def __attrs_post_init__(self):
+        """
+        Post-init hook that generates metadata for the Open edX Event.
+        """
+        # Have to use this to get around the fact that the class is frozen
+        # (which we almost always want, but not while we're initializing it).
+        # Taken from edX Learning Sequences data file.
+        object.__setattr__(self, "id", uuid1())
+        object.__setattr__(
+            self,
+            "source",
+            "openedx/{service}/web".format(
+                service=getattr(settings, "SERVICE_VARIANT", "")
+            ),
+        )
+        object.__setattr__(self, "sourcehost", socket.gethostname())
+        object.__setattr__(self, "time", datetime.utcnow())
+        object.__setattr__(
+            self, "sourcelib", tuple(map(int, openedx_events.__version__.split(".")))
+        )
--- a/openedx_events/tests/test_tooling.py
+++ b/openedx_events/tests/test_tooling.py
@@ -1,0 +1,193 @@
+"""This file contains all test for the tooling.py file.
+
+Classes:
+    EventsToolingTest: Test events tooling.
+"""
+from unittest.mock import Mock, patch
+
+import attr
+import ddt
+from django.test import TestCase, override_settings
+
+from openedx_events.exceptions import InstantiationError, SenderValidationError
+from openedx_events.tooling import OpenEdxPublicSignal
+
+
+@ddt.ddt
+class OpenEdxPublicSignalTest(TestCase):
+    """
+    Test cases for Open edX events base class.
+    """
+
+    def setUp(self):
+        """
+        Setup common conditions for every test case.
+        """
+        super().setUp()
+        self.event_type = "org.openedx.learning.session.login.completed.v1"
+        self.user_mock = Mock()
+        self.data_attr = {
+            "user": Mock,
+        }
+        self.public_signal = OpenEdxPublicSignal(
+            event_type=self.event_type,
+            data=self.data_attr,
+        )
+
+    def test_string_representation(self):
+        """
+        This methods checks the string representation for events base class.
+
+        Expected behavior:
+            The representation contains the event_type.
+        """
+        self.assertIn(self.event_type, str(self.public_signal))
+
+    @override_settings(SERVICE_VARIANT="lms")
+    @patch("openedx_events.data.openedx_events")
+    @patch("openedx_events.data.socket")
+    def test_get_signal_metadata(self, socket_mock, events_package_mock):
+        """
+        This methods tests getting the generated metadata for an event.
+
+        Expected behavior:
+            Returns the metadata containing information about the event.
+        """
+        events_package_mock.__version__ = "0.1.0"
+        socket_mock.gethostname.return_value = "edx.devstack.lms"
+        expected_metadata = {
+            "event_type": self.event_type,
+            "minorversion": 0,
+            "source": "openedx/lms/web",
+            "sourcehost": "edx.devstack.lms",
+            "sourcelib": [0, 1, 0],
+        }
+
+        metadata = self.public_signal.generate_signal_metadata()
+
+        self.assertDictContainsSubset(expected_metadata, attr.asdict(metadata))
+
+    @ddt.data(
+        ("", {"user": Mock()}, "event_type"),
+        ("org.openedx.learning.session.login.completed.v1", None, "data"),
+    )
+    @ddt.unpack
+    def test_event_instantiation_exception(
+        self, event_type, event_data, missing_argument
+    ):
+        """
+        This method tests when an event is instantiated without event_type or
+        event data.
+
+        Expected behavior:
+            An InstantiationError exception is raised.
+        """
+        exception_message = "InstantiationError {event_type}: Missing required argument '{missing_argument}'".format(
+            event_type=event_type, missing_argument=missing_argument
+        )
+
+        with self.assertRaisesMessage(InstantiationError, exception_message):
+            OpenEdxPublicSignal(event_type=event_type, data=event_data)
+
+    @patch("openedx_events.tooling.OpenEdxPublicSignal.generate_signal_metadata")
+    @patch("openedx_events.tooling.Signal.send")
+    def test_send_event_successfully(self, send_mock, fake_metadata):
+        """
+        This method tests the process of sending an event.
+
+        Expected behavior:
+            The event is sent as a django signal.
+        """
+        expected_metadata = {
+            "some_data": "data",
+            "raise_exception": True,
+        }
+        fake_metadata.return_value = expected_metadata
+
+        self.public_signal.send_event(user=self.user_mock)
+
+        send_mock.assert_called_once_with(
+            sender=None,
+            user=self.user_mock,
+            metadata=expected_metadata,
+        )
+
+    @patch("openedx_events.tooling.OpenEdxPublicSignal.generate_signal_metadata")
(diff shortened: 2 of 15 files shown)


openedx_events/tooling.py
```diff
@@ -0,0 +1,177 @@
...
+                'sourcehost': None,
+                'specversion': '0.1.0'
+            }
+        """
+
+        def get_current_time():
+            """
+            Getter function used to get timestamp when the event ocurred.
+            """
+            return str(datetime.utcnow().isoformat()) + "Z"
```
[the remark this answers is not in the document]
mariajgrimaldi: Because of this specification: https://open-edx-proposals.readthedocs.io/en/latest/oep-0041-arch-async-server-event-messaging.html#id6

openedx_events/tooling.py
```diff
@@ -0,0 +1,181 @@
...
+                    )
+                if not isinstance(argument, value):
+                    raise SenderValidationError(
+                        event_type=self.event_type,
+                        message="The argument '{key}' is not instance of the Class Attribute '{attr}'".format(
+                            key=key, attr=value.__class__.__name__
+                        ),
+                    )
+
+        validate_sender()
```
mariajgrimaldi, line 160: Hey @felipemontoya! I hope you're doing great.

I found an interesting case while implementing the POC with the 8 events. When the platform sends the signal [UNENROLL_DONE](https://github.com/edx/edx-platform/blob/63a9327a9dd6625e524c7620025225e76ac6d984/common/djangoapps/student/models.py#L1421), sends an argument called `skip_refund`. I thought it was not a big deal, I could send the argument as a keyword... but the problem is this validation -that was discussed in the [naming ADR](https://github.com/eduNEXT/openedx-events/pull/4#discussion_r636649519)-:

> We expect send calls to always send things over as keyword args, and throw an exception if we get any parameters that we don't expect, or any parameters that are missing, based on the constructor call to OpenEdxPublicSignal.

This prevents sending extra arguments not defined in the event definition. And `skip_refund` is not part of the CourseEnrollment object so it doesn't make sense to include it in the CourseEnrollmentrData class.

Then what can we do? I was thinking about just checking that `at least the arguments defined when instantiating the event are sent`, what do you think? This way we can actually send `sender` when needed.

openedx_events/tooling.py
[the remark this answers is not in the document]
felipemontoya: Adding the Z at the end of isoformat might be the easiest way but it does look like a workaround manipulation.

Even `datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%S.%fZ")` looks better imho

openedx_events/tooling.py
felipemontoya: My understanding of the spirit of the discussion we had in the naming ADR is that we precisely dont want to permit extra params. We want to make the events super explicit in what they are passing around.

In my opinion we have 2 options. We add skip_refund to make the event we are creating for unenroll an make it 100% compatible to the current `unenroll_done` or we don't and we make an event that is not 100% compatible. This means we will add it to the core platform and it will co-exist with the current unenroll_done for a while until we can migrate things.

I'd be inclined to say that we should leave without skip_refund and send two events when skip_refund is false. One for the unenrollment and one for the refund.

openedx_events/tooling.py
```diff
@@ -0,0 +1,177 @@
...
+            Getter function used to get physical source of the event.
+            """
+            current_request = crum.get_current_request()
+            return current_request.get_host() if current_request else None
+
+        def get_spec_version():
+            """
+            Getter function used to get the Open edX Events version.
+            """
+            return openedx_events.__version__
```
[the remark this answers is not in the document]
felipemontoya: I like the idea of sending the version of the openedx_events library, but I would not like to mix it with the specversion of CloudEvents.
I'd rather leave `specversion` alone as "1.0" to not mess around and send another field with the openedx events version.

For the new field I'm thinking `events_version`, but it could be other key

openedx_events/tests/test_tooling.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_send_robust_event_successfully(self, send_robust_mock):
+        """
+        This method tests the process of sending an event.
+
+        Expected behavior:
+            The event is sent as a django signal.
+        """
+        self.public_signal.send_event(user=Mock())
+
+        send_robust_mock.assert_called()
```
mariajgrimaldi, line 109: assert called with the arguments sent

openedx_events/tooling.py
mariajgrimaldi: Do you mean something like:

```
UNENROLL_DONE_WITH_REFUND
UNENROLL_DONE_WITHOUT_REFUND
```

And the receivers would be subscribed to both signals, right? I like the idea, but what if more extra arguments are needed in the future? How many variations will we need?

felipemontoya: I mean something like:

UNENROLL_DONE
UNENROLL_REFUND

openedx_events/tooling.py
```diff
@@ -0,0 +1,199 @@
...
+            )
+
+        def get_source_host():
+            """
+            Getter function used to get physical source of the event.
+            """
+            if not crum:
+                return 'Unknown host'
+            current_request = crum.get_current_request()
+            return current_request.get_host() if current_request else None
```
ormsbee, line 93: This gets the IP of the requester, but the [sourcehost](https://open-edx-proposals.readthedocs.io/en/latest/oep-0041-arch-async-server-event-messaging.html#sourcehost-extension) is meant to capture the address of the host emitting the event (i.e. `socket.gethostname()`).

openedx_events/tooling.py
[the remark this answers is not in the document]
ormsbee: You could also play off the `source*` convention and call it `sourcelib` or something, but I'm totally fine with `events_version` as well.

openedx_events/tooling.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+
+    def __init__(self, event_type, data, minor_version="0.0"):
+        """
+        Init method for OpenEdxPublicSignal definition class.
+
+        Arguments:
+            event_type (str): name of the event.
+            data (dict): attributes passed to the event.
+            minor_version (str): version of the event type.
```
ormsbee, line 31: Shouldn't the minor version be an int?

openedx_events/tooling.py
```diff
@@ -0,0 +1,199 @@
...
+        self.minor_version = minor_version
+        super().__init__()
+
+    def __repr__(self):
+        """
+        Represent OpenEdxPublicSignal as a string.
+        """
+        return "<OpenEdxPublicSignal: {event_type}>".format(event_type=self.event_type)
+
+    def get_signal_metadata(self):
```
ormsbee, line 52: A few comments:

**Defining Metadata in data.py**
Can the metadata itself be another `attrs` class that has well defined fields? This could live somewhere independently of the learning events.

**Separating CloudEvents serialization specifics.**
Let's keep the metadata here in a more natural Python form while it's in memory, and convert it to the serialized version in an explicit CloudEvents layer in the future, when that becomes relevant. In concrete terms, I think that just means that `specversion` goes away at this layer (it's a fixed constant for CloudEvents), time becomes a UTC `datetime` instead of a string, and events version might become a tuple of ints so that version comparisons don't have to worry about lexical ordering of strings (e.g. '0.9.0' vs. '0.10.0').

**Message Queue version**
_This isn't something I think you need to do yet_, but is the long term plan to make `metadata` an optional param that's auto-generated in the case of a local send, but populated from a serialized message if the signal is transmitted over a message queue?

**UUID?**
The CloudEvent will eventually have a UUID. Is it useful to have it in this metadata as well?

**Nit: "get" vs. "create/generate"**
This is a nit, but for me, `get` implies something is more or less staying the same and you're accessing it, while `create` or `generate` means that you're creating a new thing and handing it back.

openedx_events/tooling.py
feanil: I think I would do the following:

UNENROLL_DONE -> this event gets sent on any unenrollment whether or not there is a refund.
REFUND -> this event gets sent if skip_refunds = false.

In the case of a refund, you still want the unenroll message for anyone that cares about enrollments.

openedx_events/tooling.py
mariajgrimaldi: Hey there @ormsbee! I hope you're doing good!

I've updated the PR with your recommendations:

**Defining Metadata in data.py**
I've created [this](https://github.com/eduNEXT/openedx-events/pull/17/files) PR showing how I would do this

**Separating CloudEvents serialization specifics.**
I followed all your recommendations here.

**Message Queue version**
I think the best thing is to consider this use case already. If a Metadata object exists, then it's not generated. What do you think? 

**UUID?**
I added an ID as part of the metadata object, based on what the OEP-41 said:
https://open-edx-proposals.readthedocs.io/en/latest/oep-0041-arch-async-server-event-messaging.html#id2


**Nit: "get" vs. "create/generate"**
Thanks for the comment! I fixed it.

openedx_events/tooling.py
```diff
@@ -0,0 +1,189 @@
...
+        self.minor_version = minor_version
+        super().__init__()
+
+    def __repr__(self):
+        """
+        Represent OpenEdxPublicSignal as a string.
+        """
+        return "<OpenEdxPublicSignal: {event_type}>".format(event_type=self.event_type)
+
+    def generate_signal_metadata(self):
```
mariajgrimaldi, line 49: Hey @felipemontoya! what do you think about changing `generate_signal_metadata` for [this](https://github.com/eduNEXT/openedx-events/pull/17/files#diff-327e7d7bba7d257820e4e319a740dbbd12556734420a7c8df0e42583ebea6792L49) version? It uses EventsMetadata as an attr class

openedx_events/tooling.py
mariajgrimaldi: Update: we fully adopted defining metadata in data.py

openedx_events/tooling.py
felipemontoya: I like the idea of keeping the in memory data in a more natural python form and I think that the attr class for events metadata makes it well defined and robust. I'm happy to see that it was adopted already.

felipemontoya (approved):
Very good work @mariajgrimaldi.

I think this PR already covered all the feedback and the current proposal for the tooling is very robust. On my opinion we are ready to merge this.

openedx_events/data.py
```diff
@@ -0,0 +1,62 @@
...
+    OEP-41: Asynchronous Server Event Message Format.
+
+    Arguments:
+        id (UUID): event identifier.
+        event_type (str): name of the event.
+        minorversion (int): version of the event type.
+        source (str): logical source of an event.
+        sourcehost (str): physical source of the event.
+        time (datetime): timestamp when the event was sent.
+        sourcelib (str): Open edX Events library version.
```
robrap, line 32: @mariajgrimaldi: Hello Maria. It seems we went with a tuple instead of a string. I'm curious why we did that. I'm working on the event bus, which is going to need to send this metadata across the wire, so I may need to pass in a string and convert it back to a tuple, which will be a bit strange, but would make the type in this comment more correct.

mariajgrimaldi: @robrap Hello there, Robert. After some digging, I found this discussion I had with dave when this PR was open: https://github.com/openedx/openedx-events/pull/7#discussion_r667294019 If you think there's room for improvement, let me know! -besides updating the docstring 😅 -

robrap: Thank you, Maria. I'm updating in this PR: https://github.com/openedx/openedx-events/pull/168/files#r1083076265

PR #8, opened 2021-06-08
feat: add data attributes definition for learning subdomain

mariajgrimaldi:
**Description:** 
This PR adds class definitions for the objects that will be sent by Open edX events, they are based in the library

**JIRA:**
https://edunext.atlassian.net/browse/PS2021-629?atlOrigin=eyJpIjoiODIwNDA5M2MxZGRlNGQ2NjhjNjJhZWIwNWZiNWNlYjUiLCJwIjoiaiJ9

**Dependencies:** 
This PR depends on #7 

**Testing instructions:**
In your Open edX devstack
1. Go to lms shell `make lms-shell`
2. Install openedx-events specifying the events definition branch:
 `pip install git+https://github.com/eduNEXT/openedx-events.git@MJG/event_data#egg=event_data` 
3. Go to the lms Django shell: `python manage.py lms shell`
4. Now, you can import the platform objects and map them to attrs classes. For example, you could do:

```
>>> from common.djangoapps.student.models import CourseEnrollment
>>> from openedx_events.learning.data import CourseEnrollmentData, UserData, UserPersonalData, UserNonPersonalData, CourseOverviewData
>>> course_enrollment = CourseEnrollment.objects.get(id=1)

*** MAPPING CourseEnrollment to CourseEnrollmentData ***
>>> course_enrollment_data = CourseEnrollmentData(
...                             user=UserData(
...                                 user_pii=(
...                                       username=course_enrollment.user.username,
...                                       email=course_enrollment.user.email,
...                                       name=course_enrollment.user.profile.name                                 
...                                  ),
...                                 user_non_pii=(
...                                       id=course_enrollment,ent.user.id,
...                                       is_active=course_enrollment.user.is_active
...                                 ),
...                             ),
...                             course=CourseOverviewData(
...                                 course_key=course_enrollment.course.id,
...                                 display_name=course_enrollment.course.display_name
...                             ),
...                             mode=course_enrollment.mode,
...                             creation_date=course_enrollment.created,
...                             is_active=course_enrollment.is_active
...                         )
>>> print(course_enrollment_data)
>>> import attr
>>> attr.asdict(course_enrollment_data)
```

**Reviewers:**
- [ ] tag reviewer 
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/openedx_events/learning/data.py
+++ b/openedx_events/learning/data.py
@@ -1,0 +1,138 @@
+"""
+Data attributes for events within the architecture subdomain `learning`.
+
+These attributes follow the form of attr objects specified in OEP-49 data
+pattern.
+"""
+from datetime import datetime
+
+import attr
+from opaque_keys.edx.keys import CourseKey
+
+
+@attr.s(frozen=True)
+class UserNonPersonalData:
+    """
+    Attributes defined for Open edX user object based on non-PII data.
+
+    Arguments:
+        id (int): unique identifier for the Django User object.
+        is_active (bool): indicates whether the user is active.
+    """
+
+    id = attr.ib(type=int)
+    is_active = attr.ib(type=bool, default=True)
+
+
+@attr.s(frozen=True)
+class UserPersonalData:
+    """
+    Attributes defined for Open edX user object based on PII data.
+
+    Arguments:
+        username (str): username associated with the Open edX user.
+        email (str): email associated with the Open edX user.
+        name (str): email associated with the Open edX user's profile.
+    """
+
+    username = attr.ib(type=str)
+    email = attr.ib(type=str)
+    name = attr.ib(type=str, factory=str)
+
+
+@attr.s(frozen=True)
+class UserData:
+    """
+    Attributes defined for Open edX user object.
+
+    Arguments:
+        user_non_pii (UserNonPersonalData): user's Personal Identifiable
+        Information.
+        user_pii (UserPersonalData): user's Non Personal Identifiable
+        Information.
+    """
+
+    user_non_pii = attr.ib(type=UserNonPersonalData)
+    user_pii = attr.ib(type=UserPersonalData)
+
+
+@attr.s(frozen=True)
+class CourseData:
+    """
+    Attributes defined for Open edX Course Overview object.
+
+    Arguments:
+        course_key (str): identifier of the Course object.
+        display_name (str): display name associated with the course.
+        start (datetime): start date for the course.
+        end (datetime): end date for the course.
+    """
+
+    course_key = attr.ib(type=CourseKey)
+    display_name = attr.ib(type=str, factory=str)
+    start = attr.ib(type=datetime, default=None)
+    end = attr.ib(type=datetime, default=None)
+
+
+@attr.s(frozen=True)
+class CourseEnrollmentData:
+    """
+    Attributes defined for Open edX Course Enrollment object.
+
+    Arguments:
+        user (UserData): user associated with the Course Enrollment.
+        course (CourseData): course where the user is enrolled in.
+        mode (str): course mode associated with the course.
+        is_active (bool): whether the enrollment is active.
+        creation_date (datetime): creation date of the enrollment.
+        created_by (UserData): if available, who created the enrollment.
+    """
+
+    user = attr.ib(type=UserData)
+    course = attr.ib(type=CourseData)
+    mode = attr.ib(type=str)
+    is_active = attr.ib(type=bool)
+    creation_date = attr.ib(type=datetime)
+    created_by = attr.ib(type=UserData, default=None)
+
+
+@attr.s(frozen=True)
+class CertificateData:
+    """
+    Attributes defined for Open edX Certificate data object.
+
+    Arguments:
+        user (UserData): user associated with the Certificate.
+        course (CourseData): course where the user obtained the certificate.
+        mode (str): course mode associated with the course.
+        grade (str): user's grade in this course run.
+        current_status (str): current certificate status.
+        previous_status (str): if available, pre-event certificate status.
+        download_url (str): URL where the PDF version of the certificate.
+        name (str): user's name.
+    """
+
+    user = attr.ib(type=UserData)
+    course = attr.ib(type=CourseData)
+    mode = attr.ib(type=str)
+    grade = attr.ib(type=str)
+    download_url = attr.ib(type=str)
+    name = attr.ib(type=str)
+    current_status = attr.ib(type=str)
+    previous_status = attr.ib(type=str, factory=str)
+
+
+@attr.s(frozen=True)
+class CohortData:
+    """
+    Attributes defined for Open edX Cohort Membership object.
+
+    Arguments:
+        user (UserData): user assigned to the group.
+        course (CourseData): course associated with the course group.
+        name (str): name of the cohort group.
+    """
+
+    user = attr.ib(type=UserData)
+    course = attr.ib(type=CourseData)
+    name = attr.ib(type=str)
--- a/openedx_events/learning/data.py
+++ b/openedx_events/learning/data.py
@@ -21,7 +21,7 @@ class UserNonPersonalData:
     """
 
     id = attr.ib(type=int)
-    is_active = attr.ib(type=bool, default=True)
+    is_active = attr.ib(type=bool)
 
 
 @attr.s(frozen=True)
@@ -41,19 +41,18 @@ class UserPersonalData:
 
 
 @attr.s(frozen=True)
-class UserData:
+class UserData(UserNonPersonalData):
     """
     Attributes defined for Open edX user object.
 
+    This class extends UserNonPersonalData to include PII data completing the
+    user object.
+
     Arguments:
-        user_non_pii (UserNonPersonalData): user's Personal Identifiable
-        Information.
-        user_pii (UserPersonalData): user's Non Personal Identifiable
-        Information.
+        pii (UserPersonalData): user's Personal Identifiable Information.
     """
 
-    user_non_pii = attr.ib(type=UserNonPersonalData)
-    user_pii = attr.ib(type=UserPersonalData)
+    pii = attr.ib(type=UserPersonalData)
 
 
 @attr.s(frozen=True)
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -1,5 +1,6 @@
 # Core requirements for using this application
 -c constraints.txt
 
-django
 attrs
+django
+edx-opaque-keys[django]
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,7 +10,15 @@ django==2.2.24
     # via
     #   -c requirements/constraints.txt
     #   -r requirements/base.in
+edx-opaque-keys[django]==2.2.2
+    # via -r requirements/base.in
+pbr==5.6.0
+    # via stevedore
+pymongo==3.11.4
+    # via edx-opaque-keys
 pytz==2021.1
     # via django
 sqlparse==0.4.1
     # via django
+stevedore==3.3.0
+    # via edx-opaque-keys
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -88,6 +88,8 @@ docutils==0.17.1
     #   readme-renderer
 edx-lint==5.0.0
     # via -r requirements/quality.txt
+edx-opaque-keys[django]==2.2.2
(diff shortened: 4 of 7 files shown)


andrey-canon (approved):

openedx_events/learning/data.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    Attributes defined for Open edX student object.
+    """
+
+    username = attr.ib(type=str)
+    email = attr.ib(type=str)
+    first_name = attr.ib(type=str, factory=str)
+    last_name = attr.ib(type=str, factory=str)
+    is_active = attr.ib(type=bool, default=True)
+    profile = attr.ib(type=UserProfileData, default=UserProfileData())
```
felipemontoya, line 34: Do we definitely want to keep StudentData and UserProfileData as separate entities?
Does it make sense to have one without the other?

This looks like an implementation detail that is just so because of the way edx-platform extends the django auth User. For other packages we could skip this and send it as a single object.

Can you think of a place where we would not want this?

openedx_events/learning/data.py
```diff
@@ -0,0 +1,88 @@
...
+class RegistrationData:
+    """
+    Attributes defined for Open edX registration profile object.
+    """
+
+    activation_key = attr.ib(type=str)
+
+
+@attr.s(frozen=True)
+class CourseOverviewData:
```
felipemontoya, line 47: Is the course overview what we want to pass around? or is it the course data directly?

felipemontoya (changes requested):

openedx_events/learning/data.py
felipemontoya: We will probably fill the data in via what is in the CourseOverview data to avoid querying mongo, but still CourseOverview seems to me like something that is too lms implementation specific

openedx_events/learning/data.py
mariajgrimaldi: Now that I thinking about it, it doesn't make sense to send just the user or just the profile. They always send the whole user.
I'll change it and add a _profile_ namespace.

Besides `meta` and `name`, do you have any other profile fields in mind?

openedx_events/learning/data.py
mariajgrimaldi: Right. I'll change it to CourseData.

Besides `course_key` and `display_name`, do you have any other course fields in mind? 

These are all we can access without querying mongo i.e from CourseOverview model:

    # Course identification
    id = CourseKeyField(db_index=True, primary_key=True, max_length=255)
    _location = UsageKeyField(max_length=255)
    org = TextField(max_length=255, default='outdated_entry') **--> SELECTED but is outdated?**
    display_name = TextField(null=True)
    display_number_with_default = TextField()
    display_org_with_default = TextField()

    start = DateTimeField(null=True) **--> SELECTED**
    end = DateTimeField(null=True) **--> SELECTED**

    # These are deprecated and unused, but cannot be dropped via simple migration due to the size of the downstream
    # history table. See DENG-19 for details.
    # Please use start and end above for these values.
    start_date = DateTimeField(null=True)
    end_date = DateTimeField(null=True)

    advertised_start = TextField(null=True)
    announcement = DateTimeField(null=True)

    # URLs
    # Not allowing null per django convention; not sure why many TextFields in this model do allow null
    banner_image_url = TextField()
    course_image_url = TextField() **--> SELECTED (can be use in advertisement?) **
    social_sharing_url = TextField(null=True) **--> SELECTED (can be use in advertisement?) **
    end_of_course_survey_url = TextField(null=True)

    # Certification data
    certificates_display_behavior = TextField(null=True)
    certificates_show_before_end = BooleanField(default=False)
    cert_html_view_enabled = BooleanField(default=False)
    has_any_active_web_certificate = BooleanField(default=False)
    cert_name_short = TextField()
    cert_name_long = TextField()
    certificate_available_date = DateTimeField(default=None, null=True)  

    # Grading
    lowest_passing_grade = DecimalField(max_digits=5, decimal_places=2, null=True) **--> SELECTED**

    # Access parameters
    days_early_for_beta = FloatField(null=True)
    mobile_available = BooleanField(default=False)
    visible_to_staff_only = BooleanField(default=False)
    _pre_requisite_courses_json = TextField()  # JSON representation of list of CourseKey strings

    # Enrollment details
    enrollment_start = DateTimeField(null=True)
    enrollment_end = DateTimeField(null=True)
    enrollment_domain = TextField(null=True)
    invitation_only = BooleanField(default=False)
    max_student_enrollments_allowed = IntegerField(null=True)

    # Catalog information
    catalog_visibility = TextField(null=True)
    short_description = TextField(null=True)
    course_video_url = TextField(null=True)
    effort = TextField(null=True)
    self_paced = BooleanField(default=False)
    marketing_url = TextField(null=True)
    eligible_for_financial_aid = BooleanField(default=True)

    # Course highlight info, used to guide course update emails
    has_highlights = NullBooleanField(default=None)  # if None, you have to look up the answer yourself

    language = TextField(null=True)

    history = HistoricalRecords()

openedx_events/learning/data.py
```diff
@@ -0,0 +1,89 @@
...
+class CertificateData(CourseEnrollmentData):
+    """
+    Attributes defined for Open edX Certificate data object.
+    """
+
+    grade = attr.ib(type=str)
+    status = attr.ib(type=str)
+    download_url = attr.ib(type=str)
+    name = attr.ib(type=str)
+    is_active = attr.ib(init=False, default=True)
```
mariajgrimaldi, line 79: should we send `verify_uuid     - Unique identifier for the certificate`?

openedx_events/learning/data.py
```diff
@@ -0,0 +1,89 @@
...
+    profile_name = attr.ib(type=str, factory=str)
+
+
+@attr.s(frozen=True)
+class RegistrationData:
+    """
+    Attributes defined for Open edX registration profile object.
+    """
+
+    activation_key = attr.ib(type=str)
```
mariajgrimaldi, line 44: This class was added to match [REGISTER_USER](https://github.com/eduNEXT/edx-platform/blob/384bc6b5147423f9c3208ce5c4afa79e5e0cd040/openedx/core/djangoapps/user_authn/views/register.py#L251) signal, but registration is only used in [these places](https://github.com/eduNEXT/edx-platform/blob/master/openedx/core/djangoapps/user_authn/views/register.py#L235-L240). I don't think `matching REGISTER_USER` is a sufficient argument to implement this class. I'll remove it.

openedx_events/learning/data.py
```diff
@@ -0,0 +1,84 @@
...
+@attr.s(frozen=True)
+class StudentData:
+    """
+    Attributes defined for Open edX student object.
+    """
+
+    username = attr.ib(type=str)
+    email = attr.ib(type=str)
+    is_active = attr.ib(type=bool, default=True)
+    meta = attr.ib(type=Dict[str, str], factory=dict)
```
ormsbee, line 23: Are the contents of `meta` important enough to be in an event? If so, is it possible to parse it out into separate fields and define those here? I think it's better to more strictly define things where we can, so people don't have to worry about what kinds of data might be in a freeform dict.

openedx_events/learning/data.py
```diff
@@ -0,0 +1,84 @@
...
+import attr
+from opaque_keys.edx.keys import CourseKey
+
+
+@attr.s(frozen=True)
+class StudentData:
+    """
+    Attributes defined for Open edX student object.
+    """
+
```
ormsbee, line 19: Is there a reason we're not passing the ID of the student?

openedx_events/learning/data.py
```diff
@@ -0,0 +1,84 @@
...
+    username = attr.ib(type=str)
+    email = attr.ib(type=str)
+    is_active = attr.ib(type=bool, default=True)
+    meta = attr.ib(type=Dict[str, str], factory=dict)
+    name = attr.ib(type=str, factory=str)
```
ormsbee, line 24: Is is possible to separate out the PII parts (email, name) of this data into a separate object from the non-PII parts? I strongly suspect we'll want to annotate this stuff eventually, but having explicit separation is something simple we can do now.

openedx_events/learning/data.py
```diff
@@ -0,0 +1,84 @@
...
+class CourseEnrollmentData:
+    """
+    Attributes defined for Open edX Course Enrollment object.
+    """
+
+    user = attr.ib(type=StudentData)
+    course = attr.ib(type=CourseData)
+    mode = attr.ib(type=str)
+    is_active = attr.ib(type=bool)
+
```
ormsbee, line 59: Missing datetime of enrollment?

ormsbee: Or is that because it's going in a separate higher-level datetime field for the event as a whole?

openedx_events/learning/data.py
```diff
@@ -0,0 +1,84 @@
...
+    course_key = attr.ib(type=CourseKey)
+    display_name = attr.ib(type=str, factory=str)
+    start = attr.ib(type=datetime, default=None)
+    end = attr.ib(type=datetime, default=None)
+
+
+@attr.s(frozen=True)
+class CourseEnrollmentData:
+    """
+    Attributes defined for Open edX Course Enrollment object.
```
ormsbee, line 52: Is there anything about the context of the enrollment that we want to capture? Like who did the enrolling?

ormsbee (commented):
A few minor comments. Please don't consider me blocking on any of these.

openedx_events/learning/data.py
```diff
@@ -0,0 +1,84 @@
...
+    name = attr.ib(type=str, factory=str)
+
+
+@attr.s(frozen=True)
+class RegistrationFormData:
+    """
+    Attributes defined for Open edX student object.
+    """
+
+    account_form = attr.ib(type=Dict[str, str], factory=dict)
```
feanil, line 33: Similar to @ormsbee 's note above.  Is this data consistent/important enough that we should parse it before we put it into an event? The flexibility of a generic dict is nice but can lead to breakage because now we have an un-versioned schema inside our nice versioned schema.

openedx_events/learning/data.py
feanil: I would maybe send the money related stuff as a separate event from events related to learning. I'm thinking of it as  a separation between things important from an ecommerce domain perspective and the teaching/learning perspective.

(Loosely held opinion)

openedx_events/learning/data.py
mariajgrimaldi: Thanks for your suggestions! We didn't find a real relevant use case where we could use this Data class, so we decided to remove it. Either way, we could review this later when we start to use the classes.

openedx_events/learning/data.py
felipemontoya: I see you put it in the UserNonPersonalData attr class. Given the "LMS user_id can be used for public events" portion of OEP-32 I think we are in the clear

openedx_events/learning/data.py
```diff
@@ -0,0 +1,138 @@
...
+    Attributes defined for Open edX user object.
+
+    Arguments:
+        user_non_pii (UserNonPersonalData): user's Personal Identifiable
+        Information.
+        user_pii (UserPersonalData): user's Non Personal Identifiable
+        Information.
+    """
+
+    user_non_pii = attr.ib(type=UserNonPersonalData)
```
felipemontoya, line 55: I'd like to keep the access to the user data a little less convoluted.
With this, to access the ID, one would need to call `user_data_obj.user_non_pii.id`

We could do something like:
```python
@attr.s(frozen=True)
class UserNonPiiData:
    id = attr.ib(type=int)
    is_active = attr.ib(type=bool)


@attr.s(frozen=True)
class UserPiiData:
    username = attr.ib(type=str)
    email = attr.ib(type=str)
    name = attr.ib(type=str, factory=str)


@attr.s(frozen=True)
class UserData(UserNonPersonalData):
    pii = attr.ib(type=UserPersonalData)


user_data_obj = UserData(
    pii=UserPiiData(
        username="course_enrollment.user.username",
        email="course_enrollment.user.email",
        name="course_enrollment.user.profile.name"
    ),
    id=1,
    is_active=True
)
```
So the access is more direct, but we still make a strong distinction for pii data.

felipemontoya: I had to remove the is_active default when testing on the interactive console

PR #16, opened 2021-07-08
fix: add constraint to django so edx-opaque-keys can be used

mariajgrimaldi:
**Description:** 

This PR fixes the following in PR #8:


requirements/constraints.txt
```diff
@@ -7,3 +7,6 @@
 # link to other information that will help people in the future to remove the
 # pin when possible.  Writing an issue against the offending project and
 # linking to it here is good.
+
+# edX opaque keys Django constraint
+django < 3.0, >= 2.2
```
andrey-canon, line 12: is the  2.2 required ? I mean this will always take the major available version

mariajgrimaldi: I'm just following the edx-opaque-keys constraints file https://github.com/edx/opaque-keys/blob/6c7b635df41bb6b91ba4d7ee4a57b314ef95d38d/requirements/constraints.txt#L19

If we use django >= 2.2 then the error will be raised again, right?

andrey-canon: yes, but if you use django < 3.0  is the same result I think the other part is useful if someone makes upgrade in python 2 ... but who do that

andrey-canon (approved):

PR #19, opened 2021-07-26
test: remove max-doc-length from tox.ini file due to community recommendation

mariajgrimaldi:
This PR adds support for 120 max doc lines. Suggestion taken from: https://github.com/eduNEXT/openedx-events/pull/18#discussion_r675183610

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ ignore=D001
 [pycodestyle]
 exclude = .git,.tox
 max-line-length = 120
-max-doc-length = 79
+max-doc-length = 120
 
 [pydocstyle]
 ; D101 = Missing docstring in public class


tox.ini
```diff
@@ -9,7 +9,6 @@ ignore=D001
 [pycodestyle]
 exclude = .git,.tox
 max-line-length = 120
-max-doc-length = 79
```
MoisesGSalas, line 12: Do you want to remove the check altogether and not go for `max-doc-length = 120`?

mariajgrimaldi: I was following edx-django-utils configuration. But then I found this: https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#syntax-and-organization

> Please use 79 chars as a limit for docstring lines though, so that the text remains readable.

I think I'm gonna close this issue. Thanks for the review though!

mariajgrimaldi: I fixed it! @MoisesGSalas

MoisesGSalas (approved):

PR #29, opened 2021-09-08
feat: change send_robust to default method

mariajgrimaldi:
**Description:** 
This PR changes which send method to use when sending an event, so they don't fail during runtime. Instead, they will be allowed to fail in a testing/dev environment.

**Testing instructions:**

Test new default:
One of the ways to test the new default besides a production environment is in a test environment where no events are turned on. Meaning, the test class has the mixin `OpenEdxEventsTestMixin` with `ENABLED_OPENEDX_EVENTS = []`. For example: 
https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/tests/test_enrollment.py#L34

1. In a testing environment, connect a handler to one of the Open edX Events available. For example, from a plugin you could do:
```
                    {
                        "receiver_func_name": "enrollment_receiver",
                        "signal_path": "openedx_events.learning.signals.COURSE_ENROLLMENT_CREATED",
                    },
```
Where:
```
def enrollment_receiver(*args, **kwargs):
    """
    Login Open edX Event receiver.
    """
    raise Exception
```
2. Run one of the tests from test_enrollment, it could be: test_unenroll. 
`pytest common/djangoapps/student/tests/test_enrollment.py::EnrollmentTest::test_unenroll`

When the test case executes, it calls CourseEnrollment.enroll, this will fire the signal that will be received by enrollment_receiver. Given that the default is send_robust, the test shouldn't fail, just logs the result:
event-response-log

If we wanted to test that the receiver fails, then add to `ENABLED_OPENEDX_EVENTS = []` the name of the event `"org.openedx.learning.course.enrollment.created.v1"`.

Test failure in dev environment:
1. In a development environment, connect a handler to one of the Open edX Events available. For example, from a plugin you could do:
```
                    {
                        "receiver_func_name": "enrollment_receiver",
                        "signal_path": "openedx_events.learning.signals.COURSE_ENROLLMENT_CREATED",
                    },
```
Where:
```
def enrollment_receiver(*args, **kwargs):
    """
    Login Open edX Event receiver.
    """
    raise Exception
```
2. Enroll in a course, the exception must appear in the LMS.

**Reviewers:**

- [x] @felipemontoya 

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:** 
None for now.

### DIFF
--- a/openedx_events/tooling.py
+++ b/openedx_events/tooling.py
@@ -2,11 +2,16 @@
 Tooling necessary to use Open edX events.
 """
 import warnings
+from logging import getLogger
 
+from django.conf import settings
 from django.dispatch import Signal
 
 from openedx_events.data import EventsMetadata
 from openedx_events.exceptions import SenderValidationError
+from openedx_events.utils import format_responses
+
+log = getLogger(__name__)
 
 
 class OpenEdxPublicSignal(Signal):
@@ -30,6 +35,7 @@ class OpenEdxPublicSignal(Signal):
         self.event_type = event_type
         self.minor_version = minor_version
         self._allow_events = True
+        self._allow_send_event_failure = False
         self.__class__.instances.append(self)
         self.__class__._mapping[self.event_type] = self
         super().__init__()
@@ -80,7 +86,7 @@ class OpenEdxPublicSignal(Signal):
             minorversion=self.minor_version,
         )
 
-    def send_event(self, send_robust=False, **kwargs):
+    def send_event(self, send_robust=True, **kwargs):
         """
         Send events to all connected receivers.
 
@@ -147,9 +153,17 @@ class OpenEdxPublicSignal(Signal):
 
         kwargs["metadata"] = self.generate_signal_metadata()
 
-        if send_robust:
-            return super().send_robust(sender=None, **kwargs)
-        return super().send(sender=None, **kwargs)
+        if self._allow_send_event_failure or settings.DEBUG or not send_robust:
+            return super().send(sender=None, **kwargs)
+
+        responses = super().send_robust(sender=None, **kwargs)
+        log.info(
+            "Responses of the Open edX Event <%s>: \n%s",
+            self.event_type,
+            format_responses(responses, depth=2),
+        )
+
+        return responses
 
     def send(self, sender, **kwargs):  # pylint: disable=unused-argument
         """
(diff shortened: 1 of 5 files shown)


openedx_events/tooling.py
```diff
@@ -147,9 +153,17 @@ def validate_sender():
...
-            return super().send_robust(sender=None, **kwargs)
-        return super().send(sender=None, **kwargs)
+        if self._allow_send_event_failure or settings.DEBUG:
+            return super().send(sender=None, **kwargs)
+
+        responses = super().send_robust(sender=None, **kwargs)
+        log.info(
+            "Responses of the Open edX Event <%s>: \n%s",
+            self.event_type,
+            pformat(responses, depth=2),
```
felipemontoya, line 163: From the image you left in the cover letter I get that we would print something like:

`function enrollment_receiver at 0xAAAAAA, Exception()`

for each response.

Can we make it log the stacktrace in case there is an exception?
Also, can we print the package or full path of the function? logging the function object does not give great information for debug

openedx_events/tooling.py
```diff
@@ -80,7 +86,7 @@ def generate_signal_metadata(self):
             minorversion=self.minor_version,
         )
 
-    def send_event(self, send_robust=False, **kwargs):
+    def send_event(self, **kwargs):
```
felipemontoya, line 89: In case the above question is not true, do we really want to stop having this capability here?

openedx_events/tooling.py
```diff
@@ -30,6 +35,7 @@ def __init__(self, event_type, data, minor_version=0):
         self.event_type = event_type
         self.minor_version = minor_version
         self._allow_events = True
+        self._allow_send_event_failure = False
```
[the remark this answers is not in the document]
mariajgrimaldi: Unfortunately, no. The `__init__ `method only accepts: 

`def __init__(self, event_type, data, minor_version=0):`

openedx_events/tooling.py
mariajgrimaldi: You're right, we shouldn't. I'll change this to: `send_robust=True,`

openedx_events/tooling.py
mariajgrimaldi: This was what I managed to do:
log-event-responses
What do you think?

openedx_events/tooling.py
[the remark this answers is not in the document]
felipemontoya: But we could add it, right? To give more flexibility ot the event caller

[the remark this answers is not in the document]
felipemontoya: Ok, I see you added it to the send_event function. So it is no longer necesary

felipemontoya (approved):
With the latests round of feedback that keeps the send_event signature unaltered + all the goodies for logging I think we are good to merge this.

PR #31, opened 2021-11-22
feat: avro and attr bridge for event bus

jinder1s:
Avro to Attr Bridge code: 
This code handles converting betwen avro and attr. It can handle handle custom types such as datetime or CourseKey if the right extension classes are passed into the bridge.

Its still unclear how this will interact with kafka schema registry.

To see how this is used, see file "lms/djangoapps/arch_experiments/event_bus/course_events.py" in [lms/djangoapps/arch_experiments/event_bus/course_events.py
](https://github.com/edx/edx-platform/pull/29486)

**JIRA:**  https://openedx.atlassian.net/browse/ARCHBOM-1944

** Testing Instructions **
make requirements
pip install attrs fastavro
pytest

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


openedx_events/avro_attrs_bridge.py
```diff
@@ -0,0 +1,203 @@
...
+        # TODO what should the default values of the following be
+        avro_record = dict(
+            id="1",
+            type="Test.v1",
+            time="uea",
+            source="test_attrs",
+            sourcehost="enki",
+            minorversion=0,
+            data=obj_as_dict,
+        )
```
jinder1s, line 151: For reviewer: these values are hardcoded in, I currently don't know what their eventual value should be. If you have suggestions, suggest ahead!

openedx_events/avro_attrs_bridge_extensions.py
```diff
@@ -0,0 +1,63 @@
...
+    def serialize(self, obj) -> bytes:
+        ...
+
+    @abstractmethod
+    def deserialize(self, data: bytes) -> object:
+        ...
+
+    @abstractmethod
+    def record_fields(self):
+        ...
```
jinder1s, line 31: For reviewer: I choose to write the code for extensions into classes. I could have chooses Named Tuples or even a dict. I am open to alternatives.

openedx_events/avro_attrs_bridge.py
```diff
@@ -0,0 +1,224 @@
...
+            attrs_cls: Attr Class Object (not instance)
+            extensions: dict mapping Class Object to its AvroAttrsBridgeExtention subclass instance
+        """
+        self._attrs_cls = attrs_cls
+        if extensions is None:
+            self.extensions = {}
+        else:
+            self.extensions = extensions
+
+        # used by record_field_for_attrs_class function to track of which records have already been defined in schema
```
robrap, line 46:
```suggestion
        # used by record_field_for_attrs_class function to keep track of which records have already been defined in schema
```

openedx_events/avro_attrs_bridge.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        # Make an valid avro record from the attrs class. in the future `ID`, and time would be generated, the rest
+        # would be defined once and maybe passed into the class at instantiation time?
+        # Not sure if it makes sense to keep version info here since the schema registry will actually
+        # keep track of versions and the topic can have only one associated schema at a time.
+        obj_as_dict = attr.asdict(obj, value_serializer=self.extension_serializer)
+        # TODO what should the default values of the following be
+        avro_record = dict(
+            id="1",
+            type="Test.v1",
```
jinder1s, line 155: @dianakhuang This is what to_dict function should return

dianakhuang:
Managed to test this using the [POC PR](https://github.com/edx/edx-platform/pull/29486) on devstack and verified that we are able to serialize and deserialize CourseKeys.

dianakhuang (commented):
I didn't manage to get as far on this as I would have liked. Even though we are putting everything through the AvroSerializer to write to Kafka, the message we're getting back is missing the magic byte and the AvroDeserializer is rejecting it. I did note some of the other bugs I found along the way, though.

openedx_events/tests/test_avro.py
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env python3
+import io
+import fastavro
+import json
+
+
+import avro.schema
+from avro.datafile import DataFileReader, DataFileWriter
+from avro.io import DatumReader, DatumWriter, BinaryEncoder, BinaryDecoder
```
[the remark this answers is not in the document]
jinder1s: This was a temp file used for testing, it is now deleted

openedx_events/avro_attrs_bridge.py
```diff
@@ -0,0 +1,246 @@
...
+            # The default value is always set to None to allow attr class to handle dealing with default values
+            # in dict_to_attrs function in this class
+            if attribute.default is not attr.NOTHING:
+                inner_field['type'] = ['null', inner_field['type']]
+                inner_field['default'] = None
+            field["type"]["fields"].append(inner_field)
+
+        return field
+
+    def to_dict(self, obj):
```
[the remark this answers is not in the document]
jinder1s: I ended up wrapping this func in kafka_wrapper class (child class at bottom of this file)

dianakhuang (approved):
Just a few minor comments

openedx_events/avro_attrs_bridge.py
```diff
@@ -0,0 +1,263 @@
...
+            attrs_cls: Attr class object
+        Returns:
+            complex dict that defines avro schema for attrs_cls
+        """
+        base_schema = {
+            "namespace": "io.cloudevents",
+            "type": "record",
+            "name": "CloudEvent",
+            "version": "1.0",
+            "doc": "Avro Event Format for CloudEvents",
```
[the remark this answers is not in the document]
jinder1s: What do you think  of Avro Event Format for CloudEvents created with openedx_events/avro_attrs_bridge?

openedx_events/avro_attrs_bridge.py
```diff
@@ -0,0 +1,273 @@
...
+        Converts obj into dictionary that matches avro schema (self.schema).
+
+        Args:
+            obj: instance of self._attr_cls
+            context: dict with following value overwrites:
+                - id: unique id for this event. If id is not in dict, a uuid1 will be created for this event
+                - time: time stamp for this event. If time is not in dict, datetime.now() will be called
+        """
+        # Not sure if it makes sense to keep version info here since the schema registry will actually
+        # keep track of versions and the topic can have only one associated schema at a time.
```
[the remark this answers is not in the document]
jinder1s: I left it there cause we don't really know how this will integrate with schema registry. In my mind, this is still an open question.

openedx_events/avro_attrs_bridge.py
```diff
@@ -0,0 +1,263 @@
...
+        obj_as_dict = attr.asdict(obj, value_serializer=self.extension_serializer)
+        # TODO what should the default values of the following be
+        avro_record = dict(
+            id=context["id"]
+            if isinstance(context, dict) and "id" in context
+            else str(uuid.uuid1()),
+            type=self.config["type"],
+            time=context["time"]
+            if isinstance(context, dict) and "time" in context
+            else datetime.now().isoformat(),
```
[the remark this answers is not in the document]
jinder1s: good idea

openedx_events/avro_attrs_bridge.py
```diff
@@ -0,0 +1,287 @@
...
+        """
+        Generate avro schema for attr_cls
+
+        Arguments:
+            attrs_cls: Attr class object
+        Returns:
+            complex dict that defines avro schema for attrs_cls
+        """
+        base_schema = {
+            "namespace": "io.cloudevents",
```
robrap, line 89: 1. I just found https://github.com/cloudevents/spec/blob/v1.0.1/kafka-protocol-binding.md. Have you read this? For example, see `ce_` prefix notes. I'm not clear on when this comes into play.
2. It looks like `time` is optional. Should we leave out any optional attributes if we are unsure? See https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#optional-attributes
3. Is `type` supposed to be something like: https://github.com/eduNEXT/openedx-events/blob/2fba7ae850ae5e0cb6f56026ec17d867bac23d94/openedx_events/learning/signals.py#L19? Are these Signal classes supposed to be for internal/external events (even though Signal is in the name), or was the intention to create new classes for these events with a similar type?
4. Are we using CloudEvent types for all event data? For example, would timestamps be handled correctly? See https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#type-system
5. Do we have any clear use case of how/where using the CloudEvent spec will be helpful? Is there a tool or anything we can use to validate that we are in fact doing what we need to be doing? I found this validator, but without a specific use case, I'm not sure that we can ensure we are doing what we need to be doing (see https://www.itb.ec.europa.eu/json/cloudevents/upload)
6. Can you add a comment with a link to the spec and to the OEP to clarify our use of CloudEvents?

Some of this may require discussion with eduNEXT and Dave O. This may require iterating. Hopefully you can help manage what is good enough and what next steps (tickets) are needed. Thanks!

robrap (commented):
Thanks Jinder. Mostly a bunch of food-for-thought regarding CloudEvents.

openedx_events/avro_attrs_bridge.py
jinder1s: This is interesting. 

For 1. I don't know where kafka-protocol-bindings would come into play for us. So far, I've only dealt with how cloudevents work with avro format.

2. I added time based on specification in [OEP-41](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0041-arch-async-server-event-messaging.html).  How editable is that OEP and how much can we deviate from it for our work?
3.  Yes, I think the type is suppose to be like that. I don't know much about those signal classes.
4. I think we are using cloudevent types for time correctly. We'll be outputting timestamps in ISO format.
5. I haven't yet found a validator for cloundevents specification.
6. I will comment with link to spec and OEP soon.

jinder1s:
The cloudevents additions to schema produced by avro_attrs_bridge is based on[ cloudevents avro spec](https://github.com/cloudevents/spec/blob/v1.0.1/avro-format.md) and [OEP-41](https://github.com/cloudevents/spec/blob/v1.0.1/avro-format.md)

PR #36, opened 2022-01-12
feat: Adding LicenseLifecycle data event

jinder1s:
Creating a TrackingEvent data type to be used with https://github.com/openedx/license-manager/pull/368

- [ ] tag reviewer 
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


openedx_events/enterprise/data.py
```diff
@@ -0,0 +1,44 @@
...
+        assigned_date (str):
+        activation_date (str):
+        assigned_lms_user_id (str):
+        assigned_email (str):
+        expiration_processed (bool):
+        auto_applied (bool):
+        enterprise_customer_uuid (str):
+        enterprise_customer_slug (str):
+        enterprise_customer_name (str):
+        customer_agreement_uuid (str):
```
jinder1s, line 30: Could you help me document this or point me to a place where I could get this info? @rgarber

openedx_events/enterprise/data.py
```diff
@@ -0,0 +1,50 @@
...
+                                    May be empty if this license is not activated yet.
+        assigned_email (str): The email assigned to this license.
+                              Will be empty if this license is not assigned to a learner yet.
+        expiration_processed (bool): Boolean value of License 'expiration_processed' field, may be True or False.
+                                     True means a license is expired.
+        auto_applied (bool):
+        enterprise_customer_uuid (str):
+        enterprise_customer_slug (str):
+        enterprise_customer_name (str):
+        customer_agreement_uuid (str): The active Customer Agreement UUID that this license is linked to.
```
jinder1s, line 35: Could I get help with filling out and validating these descriptions?

macdiesel: enterprise_customer_uuid (str): UUID used to internally represent an enterprise 
enterprise_customer_slug (str): Short name of an enterprise used in the URLs and other places for display purposes
enterprise_customer_name (str): Long Name of an enterprise customer. Used for display and reporting.

openedx_events/enterprise/data.py
```diff
@@ -0,0 +1,53 @@
...
+Data attributes for events within the architecture subdomain `enterprise`.
+
+These attributes follow the form of attr objects specified in OEP-49 data
+pattern.
+"""
+import attr
+
+
+@attr.s(frozen=True)
+class TrackingEvent:
```
robrap, line 11: This should be a much more specific name, right?

jinder1s: How about LicenseLifeCycle? Based on: https://github.com/openedx/license-manager/blob/d298a722d292ba91d4b1c8afa88c4bd927752492/docs/segment_events.rst#event-nested-data-field-specification

dianakhuang (approved):

openedx_events/enterprise/data.py
```diff
@@ -0,0 +1,54 @@
...
+Data attributes for events within the architecture subdomain `enterprise`.
+
+These attributes follow the form of attr objects specified in OEP-49 data
+pattern.
+"""
+import attr
+
+
+@attr.s(frozen=True)
+class LicenseLifecycle:
```
robrap, line 11: What do you think of `SubscriptionLicenseData`?
1. See separate PR comments around "lifecycle".
2. I think there is a pattern of using `Data` as a suffix for these objects.

openedx_events/enterprise/data.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Event to track changes to subscriptsion.License records.
```
robrap, line 13: 1. typo: subscriptsion => subscription
2. Why not "subscription licenses"?
3. Other data objects use a description more like: "Attributes defined for Open edX Subscription License object."

openedx_events/enterprise/signals.py
```diff
@@ -0,0 +1,23 @@
...
+conventions specified in docs/decisions/0002-events-naming-and-versioning.rst
+
+They also must comply with the payload definition specified in
+docs/decisions/0003-events-payload.rst
+"""
+
+from openedx_events.enterprise.data import LicenseLifecycle
+from openedx_events.tooling import OpenEdxPublicSignal
+
+# .. event_type: org.openedx.enterprise.subscription.license.modified.v1
```
robrap, line 14: Modified or created? Not in sync with the rest of the code.

openedx_events/enterprise/signals.py
```diff
@@ -0,0 +1,23 @@
...
+# .. event_name: LICENSE_CREATED
+# .. event_description: emitted with Subscription.License is created.
```
robrap, line 16: 1. with => when?
2. Can we use plain english, rather than dotted notation (e.g. "Subscription License")?
3. Can you add to the description on its own line, something like: "WARNING: This event is being used for prototype purposes and should not be used, and should not be copied as a fully vetted pattern."
4.

openedx_events/enterprise/signals.py
```diff
@@ -0,0 +1,23 @@
...
+# .. event_data: LicenseLifecycle
+LICENSE_CREATED = OpenEdxPublicSignal(
+    event_type="org.openedx.enterprise.subscription.license.created.v1",
```
robrap, line 19: Given the "warning" I asked to have added in a separate comment, I don't think this comment is a blocker, but I'd like to start a discussion here.
1. The "lifecycle" part of the data was moved to the event_type and name. That is why I am proposing it shouldn't really be in the Data object name. At least, according to patterns established by things like course enrollment events: https://github.com/eduNEXT/openedx-events/blob/main/openedx_events/learning/signals.py#L38-L59
2. We need to research a bit more about how to handle the translation of this to both topics and consumer. The book we were reading was recommending one event type per topic, and one topic per consumer. During book club discussion, for lifecycle events, some people were assuming that we might follow that pattern, but that "created" vs "revoked" etc. would be part of the data, which would make it simpler to ensure proper ordering, rather than working across topics. Also, handling multiple lifecycle topics across multiple consumers would add even more complexity. This may not be clear, other than to say we need to spend more time on the best pattern for this.
3. Lastly, is it possible we may want to have a different pattern for these event bus events than for signals? For example with enrollments? If so, what are the implications for what we are designing?

As stated above, I don't want this comment to be a blocker, but I think it will need more discovery/conversations. It's great that this PR can kickstart this work.

openedx_events/enterprise/signals.py
```diff
@@ -0,0 +1,23 @@
...
+    data={
+        "lifecycle": LicenseLifecycle,
```
robrap, line 21: I imagine we may want to adjust once we settle on names.

robrap (commented):
@jinder1s: Thanks for this. I know these comments are coming post-merge, but I think that is fine if we are willing to iterate quickly, given that this is all proof-of-concept work.

openedx_events/enterprise/signals.py
robrap: @jinder1s: I'm going to move 2 and 3 into a discovery ticket.

PR #49, opened 2022-03-25
build: Move where we set the DJANGO_SETTINGS_MODULE.

feanil:
It was previously being set in the tox.ini which was fine when we were building
locally but readthedocs.org does not use it.  Moving the setting into
docs/conf.py means it should work from anywhere.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,6 +42,9 @@ sys.path.append(REPO_ROOT)
 
 VERSION = get_version('../openedx_events', '__init__.py')
 
+# Set the DJANGO_SETTINGS_MODULE if it's not set.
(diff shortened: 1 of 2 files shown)


robrap (commented):
I'm not clear on what the right solution is. Happy to approve, but is there a reason this doesn't more closely match the cookiecutter? https://github.com/openedx/edx-cookiecutters/search?q=DJANGO_SETTINGS_MODULE

feanil:
Sounds like a shortcoming of the cookiecutter, I think this is a reasonable solution and common in a few blog posts I looked at so I think it's a fine fix forward.  I don't think we need to make changes to the cookiecutter since it's not guaranteed that all libraries will need to reference Django.

Issue #38, opened 2022-01-20
What to send over the wire (kafka)?


jinder1s:
Currently, arch-bom, a team at edX, is testing MVP Kafka event bus integration into the Open edX platform. During the implementation, a couple of questions were raised and github issues seems like a great public venue to have the discussion.

Questions:
1. What is sent over the over the wire in Kafka? 
    The options are:
           a. The serialized versions of the attrs decorated classes implemented in[ data.py](https://github.com/eduNEXT/openedx-events/blob/main/openedx_events/learning/data.py)
           b. The serialized version of signal classes implemented in [signal.py](https://github.com/eduNEXT/openedx-events/blob/main/openedx_events/learning/signals.py)
           c. The serialized version of data attribute in signal classes.[ Example](https://github.com/eduNEXT/openedx-events/blob/main/openedx_events/learning/signals.py#L32-L34)
2. What is sent by a django signals sender?
    I do not know django signals very well, so I curious if the intention is for django signals to send the signals classes implemented in [signal.py](https://github.com/eduNEXT/openedx-events/blob/main/openedx_events/learning/signals.py). Or will the signal sender also send exactly the same information as a Kafka producer.
3. Depending on which option we choose above for Kafka, what happens when we need to change data definition(schema)? Specifically, which unit changes? 
    Choices:
        a. Attrs decorated classes in data.py
        b. signals class in signal.py
        c. Both??

My main goal for this discussion is pinpointing what to use to create a schema for Kafka. So far, I've been using the Attrs decorated classes in data.py.

jinder1s:
@robrap I created this issue to further along discussions on Kafka/schema/signals from this convo: https://edx-internal.slack.com/archives/C024JPE27PV/p1642175923001100

ormsbee:
@mariajgrimaldi and @felipemontoya are closer to this work, but I'll take a stab at this based on some early design thoughts and my understanding of where we're at. I'm also going to answer in a slightly different order because I think it makes it easier:

## In-process Signals

> 2. What is sent by a django signals sender?
>  I do not know django signals very well, so I curious if the intention is for django signals to send the signals classes implemented in [signals.py](https://github.com/eduNEXT/openedx-events/blob/main/openedx_events/learning/signals.py). Or will the signal sender also send exactly the same information as a Kafka producer.

[...]

PR #54, opened 2022-04-28
feat: handle optional fields in bridge

rgraber:
**Description:** 
Update the AvroAttrsBridge to handle nullable/optional fields in Attrs classes

**JIRA:** [ARCHBOM-2038](https://openedx.atlassian.net/browse/ARCHBOM-2038)

**Reviewers:**
- [ ] tag reviewer 
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.


openedx_events/bridge/tests/test_utilities.py
```diff
@@ -91,6 +91,16 @@ class TestData:
     sub_test_1: SubTestData1
 
 
+@attr.s(frozen=True)
+class SimpleAttrsWithDefaults:
+    """Test attrs with optional (default=None) values"""
+    boolean_field = attr.ib(type=bool, default=None)
```
robrap, line 97: Quick initial comments:
1. See https://github.com/openedx/openedx-events/issues/38#issuecomment-1021858331 (and other notes in the issue) regarding optional vs default.
2. Is this something we need to deal with now? It may be, and it may be that we need to deal with this sooner rather than later. How will this relate to OpenEdxPublicSignal data validation?

This might warrant its own ticket, and we should think about: a) is this the right time to solve (maybe?), and b) is there a hack that could work for license manager events, especially since we don't even care about this data?

rgraber: We do need to deal with nullable fields, even if we don't call them optional. LicenseEvents have a bunch of things that may be null or may be strings, and right now the bridge can't handle that. I can change the language around the comments and methods to get rid of the idea of optionality but I do think this PR successfully handles nullable fields and so we should push it through if there are no major issues.

robrap (commented):
I had a pending comment that I didn’t realize was pending. This was written before your latest comment.

Also, I think optional field will ultimately be needed. I just want to do it right and when it is necessary. I think it will entail adjustments to the declaration that may require conversations with the community. I don’t think a quick hack will be useful here, and I’m not clear on this, and don’t want to get distracted by this.

docs/decisions/0007-optional-fields.rst
```diff
@@ -0,0 +1,44 @@
...
+parameter, then the user does not have to set the value at all. However, it seems the ``default`` parameter may not be
+fully-supported with Kafka (see https://github.com/confluentinc/kafka-rest/issues/427 ). The Avro documentation,
+similar to that of ``attrs``, uses the default parameter to define optionality.
+
+Decision
+--------
+
+- The bridge will convert ``attrs`` fields with a default value of None to Avro fields with type = ["null",<type>]
+and default = "null".
+- Top-level fields in event data definitions will not be optional. This is contrary to and overrides the decision made in
```
robrap, line 28: I need to look more closely at this. I don’t think we are overriding a decision around the top-level fields. I think we simply haven’t implemented any capability around that, which was also true before.

openedx_events/__init__.py
```diff
@@ -5,4 +5,4 @@
 more information about the project.
 """
 
-__version__ = "0.8.3"
+__version__ = "0.8.4"
```
robrap, line 8: Seems like 0.9.

openedx_events/bridge/avro_attrs_bridge.py
```diff
@@ -126,6 +122,7 @@ def _create_avro_field_definition(self, data_key, data_type):
                 " be one of the types in PYTHON_TYPE_TO_AVRO_MAPPING, an attrs decorated class, or one of the types"
                 " defined in self.extensions."
             )
+        return field
```
robrap, line 125: Consider a single if that converts the type to a Union with “null” here instead. It may be cleaner if it isn’t missing any odd cases.

robrap (commented):
Thanks.

docs/decisions/0007-optional-fields.rst
rgraber: there were some capabilities around it before the previous PR that changed the bridge to work with entire signals, but they were removed in favor of getting the basics of the bridge out faster.

docs/decisions/0007-optional-fields.rst
```diff
@@ -0,0 +1,44 @@
...
+The Avro documentation describes somewhat similar mechanics for its fields. If a field's ``type`` list includes
+"null," the user can set the value of the field to null/None. Separately, if a field is defined with a ``default``
+parameter, then the user does not have to set the value at all. However, it seems the ``default`` parameter may not be
+fully-supported with Kafka (see https://github.com/confluentinc/kafka-rest/issues/427 ). The Avro documentation,
+similar to that of ``attrs``, uses the default parameter to define optionality.
+
+Decision
+--------
+
+- The bridge will convert ``attrs`` fields with a default value of None to Avro fields with type = ["null",<type>]
```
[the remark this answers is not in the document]
rgraber: Yes, didn't want to contemplate that for this particular issue

CHANGELOG.rst
```diff
@@ -16,6 +16,12 @@ Unreleased
 Changed
 ~~~~~~~
 
+[0.9.0] - 2022-04-28
+--------------------
+Changed
+~~~~~~~
+* Updated AvroAttrsBridge to handle nullable fields
```
robrap, line 23: Nit
```suggestion
* Updated AvroAttrsBridge to handle optional/nullable fields
```

docs/decisions/0007-optional-fields.rst
```diff
@@ -0,0 +1,41 @@
...
+------------
+
+- The notions of "optionality" and "nullability" will be effectively conflated. This means no distinctions will be made between fields for which ``None`` is a valid value and fields that simply do not need to be supplied.
+- Attr fields with defaults that are not ``None`` will be treated as required
+- At the point of writing this ADR, it is unclear exactly how Kafka will treat the addition and removal of optional fields for the purposes of schema evolution (because of the issue linked above). This will require a little more testing.
+
+Deferred/Rejected Decisions
+---------------------------
+
+- We are deferring any work to around making top-level fields in event data optional. This reverses the plan laid out in ":doc:`0006-event-schema-serialization-and-evolution`", which stated that the initial bridge would treat all new top-level fields as optional and deferred any work around making sub-fields optional
```
robrap, line 40: 1. I don't think ADR 6 means to say what you are reading. Maybe we need to fix some wording there? I adjusted the text in this ADR as I think it should read.
2. Fixed typo "work to around".
```suggestion
- Top-level fields will continue to be required, as stated in ":doc:`0006-event-schema-serialization-and-evolution`". We continue to defer the possibility of making these optional.
```

openedx_events/bridge/avro_attrs_bridge.py
```diff
@@ -139,11 +139,12 @@ def _generate_avro_record_for_attrs_class(
         """
         field: Dict[str, Any] = {}
         field["name"] = field_name
-        field["type"] = dict(name=attrs_class.__name__, type="record", fields=[])
+        record_type = dict(name=attrs_class.__name__, type="record", fields=[])
+        field["type"] = record_type
```
robrap, line 143: Nit: My personal preference would be to move this line right before you return field, and after you have manipulated record_type. Just a preference though.

openedx_events/bridge/tests/test_utilities.py
robrap: ADR captures our decision. Marking as resolved.

robrap (commented):
Nice. Thank you.

docs/decisions/0007-optional-fields.rst
```diff
@@ -0,0 +1,41 @@
...
+--------
+
+- The bridge will convert ``attrs`` fields with a default value of None to Avro fields with type = ["null",<type>] and default = "null".
+
+Consequences
+------------
+
+- The notions of "optionality" and "nullability" will be effectively conflated. This means no distinctions will be made between fields for which ``None`` is a valid value and fields that simply do not need to be supplied.
+- Attr fields with defaults that are not ``None`` will be treated as required
+- At the point of writing this ADR, it is unclear exactly how Kafka will treat the addition and removal of optional fields for the purposes of schema evolution (because of the issue linked above). This will require a little more testing.
```
robrap, line 35: I’m not sure if these notes are useful for this ADR, but I’ll let you decide. I know we still need to prove out schema evolution doing what we need it to.

robrap (approved):
Nice. Thank you.

PR #61, opened 2022-06-05
fix: need the a

e0d:
**Description:** Describe in a couple of sentences what this PR adds

In order for Backstage to discover this file it needs to be yaml, not yml.


e0d:
I'm merging this as this file is only used be Backstage which I am in the process of getting running.  This PR simply renames the catalog file, updating the extension from yml -> yaml.

PR #68, opened 2022-06-21
unit tests for serializer

whuang1202:
Updated unit tests for serializers, deserializer still work in progress

### DIFF
--- a/openedx_events/event_bus/avro/tests/test_avro.py
+++ b/openedx_events/event_bus/avro/tests/test_avro.py
@@ -1,6 +1,12 @@
 """Test interplay of the various Avro helper classes"""
+from datetime import datetime
 from unittest import TestCase
 
+from opaque_keys.edx.keys import CourseKey
+
+from openedx_events.enterprise import signals  # pylint: disable=unused-import
+from openedx_events.event_bus.avro.deserializer import AvroSignalDeserializer
+from openedx_events.event_bus.avro.serializer import AvroSignalSerializer
 from openedx_events.event_bus.avro.tests.test_utilities import (
     EventData,
     NestedAttrsWithDefaults,
@@ -11,10 +17,67 @@ from openedx_events.event_bus.avro.tests.test_utilities import (
     deserialize_bytes_to_event_data,
     serialize_event_data_to_bytes,
 )
+# Each new folder with signals must be manually imported in order for the signals to be cached
+#   and used in the unit tests. Using 'disable=reimported' with pylint will work,
+#   because we just use the cached signal list.
+# TODO (EventBus): Find a way to auto-discover these signals.
+from openedx_events.learning import signals  # pylint: disable=reimported
+from openedx_events.tests.utils import FreezeSignalCacheMixin
+from openedx_events.tooling import OpenEdxPublicSignal
+
+# If a signal is explicitly not for use with the event bus, add it to this list
+#  and document why in the event's annotations
+KNOWN_UNSERIALIZABLE_SIGNALS = ["org.openedx.learning.discussions.configuration.changed.v1"]
 
 
-class TestAvro(TestCase):
+def generate_test_event_data_for_data_type(data_type):
+    """
+    Generates test data for use in the event bus test cases.
+
+    Builds data by filling in dummy data for basic data types (int/bool/str)
+    and recursively breaks down the classes for nested classes into basic data types.
+
+    Arguments:
+        data_type: The type of the data which we are generating data for
+
+    Returns:
+        (dict): A data dictionary containing dummy data for all attributes of the class
+    """
+    data_dict = {}
+    defaults_per_type = {
+        int: 1,
+        bool: True,
+        str: "default",
+        CourseKey: CourseKey.from_string("course-v1:edX+DemoX.1+2014"),
+        datetime: datetime.now(),
+    }
+    for attribute in data_type.__attrs_attrs__:
+        result = defaults_per_type.get(attribute.type, None)
+        if result is not None:
+            data_dict.update({attribute.name: result})
+        elif attribute.type in [dict, list]:
+            raise Exception("Unable to generate Avro schema for dict or array fields")
+        else:
+            data_dict.update({attribute.name: attribute.type(**generate_test_event_data_for_data_type(attribute.type))})
+    return data_dict
+
+
+class TestAvro(FreezeSignalCacheMixin, TestCase):
     """Tests for end-to-end serialization and deserialization of events"""
+    def test_all_events(self):
+        for signal in OpenEdxPublicSignal.all_events():
+            if signal.event_type in KNOWN_UNSERIALIZABLE_SIGNALS:
+                continue
+            test_data = {}
+            serializer = AvroSignalSerializer(signal)
+            for key, curr_class in signal.init_data.items():
+                example_data = generate_test_event_data_for_data_type(curr_class)
+                example_data_processed = curr_class(**example_data)
+                test_data.update({key: example_data_processed})
+            serialized = serializer.to_dict(test_data)
+            deserializer = AvroSignalDeserializer(signal)
+            deserialized = deserializer.from_dict(serialized)
+            self.assertDictEqual(deserialized, test_data)
 
     def test_full_serialize_deserialize(self):
         SIGNAL = create_simple_signal({"test_data": EventData})
--- a/openedx_events/event_bus/avro/tests/test_deserializer.py
+++ b/openedx_events/event_bus/avro/tests/test_deserializer.py
@@ -16,9 +16,10 @@ from openedx_events.event_bus.avro.tests.test_utilities import (
     SubTestData1,
     create_simple_signal,
 )
+from openedx_events.tests.utils import FreezeSignalCacheMixin
 
 
-class TestAvroSignalDeserializer(TestCase):
+class TestAvroSignalDeserializerCache(TestCase, FreezeSignalCacheMixin):
     """Test AvroSignalDeserializer"""
 
     def setUp(self) -> None:
--- a/openedx_events/event_bus/avro/tests/test_serializer.py
+++ b/openedx_events/event_bus/avro/tests/test_serializer.py
@@ -1,8 +1,8 @@
 """Tests for avro.serializer module."""
 
 from datetime import datetime
-from unittest import TestCase
 
+from django.test import TestCase
 from opaque_keys.edx.keys import CourseKey
 
 from openedx_events.event_bus.avro.serializer import AvroSignalSerializer
@@ -17,9 +17,10 @@ from openedx_events.event_bus.avro.tests.test_utilities import (
     SubTestData1,
     create_simple_signal,
 )
+from openedx_events.tests.utils import FreezeSignalCacheMixin
 
 
-class TestAvroSignalSerializer(TestCase):
+class TestAvroSignalSerializerCache(FreezeSignalCacheMixin, TestCase):
     """Tests for AvroSignalSerializer."""
 
     def test_convert_event_data_to_dict(self):
--- a/openedx_events/learning/signals.py
+++ b/openedx_events/learning/signals.py
@@ -128,6 +128,7 @@ COHORT_MEMBERSHIP_CHANGED = OpenEdxPublicSignal(
 # .. event_type: org.openedx.learning.discussions.configuration.changed.v1
 # .. event_name: COURSE_DISCUSSIONS_CHANGED
 # .. event_description: emitted when the configuration for a course's discussions changes in the course
+#       Warning: This event is currently incompatible with the event bus, list/dict cannot be serialized yet
 # .. event_data: CourseDiscussionConfigurationData
 COURSE_DISCUSSIONS_CHANGED = OpenEdxPublicSignal(
     event_type="org.openedx.learning.discussions.configuration.changed.v1",
--- a/openedx_events/tests/test_tooling.py
+++ b/openedx_events/tests/test_tooling.py
@@ -10,11 +10,12 @@ import ddt
 from django.test import TestCase, override_settings
 
 from openedx_events.exceptions import SenderValidationError
+from openedx_events.tests.utils import FreezeSignalCacheMixin
 from openedx_events.tooling import OpenEdxPublicSignal
 
 
 @ddt.ddt
-class OpenEdxPublicSignalTest(TestCase):
+class OpenEdxPublicSignalTestCache(FreezeSignalCacheMixin, TestCase):
     """
     Test cases for Open edX events base class.
     """
--- a/openedx_events/tests/utils.py
+++ b/openedx_events/tests/utils.py
@@ -1,7 +1,31 @@
 """
-Utils used by Open edX events.
+Utils used by Open edX event tests.
 """
 from openedx_events.tooling import OpenEdxPublicSignal
+
+
+class FreezeSignalCacheMixin:
+    """
+    A mixin to be used by TestCases to avoid new signals persisting in the OpenEdxPublicSignal cache of instances.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        Save current signal instances.
+        """
+        super().setUpClass()
+        cls.pre_run_instances = list(OpenEdxPublicSignal.instances)
+        cls.pre_run_mapping = dict(OpenEdxPublicSignal._mapping)  # pylint: disable=protected-access
+
+    @classmethod
+    def tearDownClass(cls):
+        """
+        Restore instance cache to pre-test state.
+        """
+        super().tearDownClass()
+        OpenEdxPublicSignal.instances = cls.pre_run_instances
+        OpenEdxPublicSignal._mapping = cls.pre_run_mapping  # pylint: disable=protected-access
 
 
 class EventsIsolationMixin:


openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +20,64 @@
...
+# The following imports are only necessary if we are running Python 3.9+:
+# import collections
+# collections.Callable = collections.abc.Callable
+
+
+
+
+
+
+def generate(class_type):
```
rgraber, line 35: rename suggestion: `generate_test_event_data_for_signal(signal)`

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +20,64 @@
...
+    """
+    Generates test data automatically by taking in the class's type
+    and breaking it down to the basic data types (int/bool/str),
+    then generates dummy data for these attributes.
+    """
+    dtdict = {}
+    for attribute in class_type.__attrs_attrs__:
+        if attribute.type == int:
```
rgraber, line 43: Strikes me as a good place for a switch statement

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +20,64 @@
...
+            dtdict.update({attribute.name: "default"})
+        elif attribute.type == CourseKey:
+            course_key = CourseKey.from_string("course-v1:edX+DemoX.1+2014")
+            dtdict.update({attribute.name: course_key})
+        elif attribute.type == datetime:
+            dtdict.update({attribute.name: datetime.now()})
+        elif attribute.type in [dict, list]:
+            raise NameError("Unable to generate Avro schema for dict or array fields")
+        else:
+            dtdict.update({attribute.name: attribute.type(**generate(attribute.type))})
```
rgraber, line 57: Cute

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +20,64 @@
...
+    return dtdict
 
 
 class TestAvro(TestCase):
     """Tests for end-to-end serialization and deserialization of events"""
+    def test_all_events(self):
+        for a in OpenEdxPublicSignal.all_events():
```
rgraber, line 64: Different variable name. Maybe `signal`?

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +18,61 @@
...
+    Recursively breaks down the classes for nested classes
+
+    Arguments:
+        class_type: The type of the class which we are generating data for
+
+    Returns:
+        dtdict: A data dictionary containing dummy data for all attributes of the class
+    """
+    dtdict = {}
+    switcher = {
```
rgraber, line 37: Nit: a more descriptive variable name would be something like `defaults_per_type`

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +18,61 @@
     deserialize_bytes_to_event_data,
     serialize_event_data_to_bytes,
 )
+from openedx_events.learning import signals  # pylint: disable=reimported
+from openedx_events.tooling import OpenEdxPublicSignal
+
+
+def generate_test_event_data_for_signal(class_type):
```
rgraber, line 25: This parameter name is misleading. The argument is not really a type, it's an instance of a signal

rgraber (changes requested):
We should also update the documentation to note that when adding new signals, if they are not to be used with the event bus, they should be added to the list of known exceptions

openedx_events/event_bus/avro/tests/test_avro.py
whuang1202: I call this function recursively for the classes contained within the signal, are the classes within the signals also themselves signals?

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,58 @@
     deserialize_bytes_to_event_data,
     serialize_event_data_to_bytes,
 )
+from openedx_events.learning import signals  # pylint: disable=reimported
+from openedx_events.tooling import OpenEdxPublicSignal
+
+KNOWN_UNSERIALIZABLE_EVENTS = ["org.openedx.learning.discussions.configuration.changed.v1"]
```
rgraber, line 23: My bad, this should probably be KNOWN_UNSERIALIZABLE_SIGNALS

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,58 @@
...
+
+def generate_test_event_data_for_signal(data_type):
```
rgraber, line 25: Also my bad, this should be `generate_test_data_for_data_type` or something since we're not actually passing it a signal

openedx_events/event_bus/avro/tests/test_avro.py
rgraber: Above comment still holds. `a` is not really descriptive

rgraber (changes requested):
We should update the documentation for adding new signals to make sure they get imported here and added to the list of exceptions if they are explicitly not meant for the event bus.

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,66 @@
...
-class TestAvro(TestCase):
+class TestAvro(FreezeSignalCacheMixin, TestCase):
     """Tests for end-to-end serialization and deserialization of events"""
+    def test_all_events(self):
+        for signal in OpenEdxPublicSignal.all_events():
+            if signal.event_type not in KNOWN_UNSERIALIZABLE_SIGNALS:
+                test_data = {}
+                serializer = AvroSignalSerializer(signal)
+                attribute_number = 0
+                for b in list(signal.init_data.keys()):
```
rgraber, line 69: I think I misled you here. I forgot that `schema.fields` is a list and not a dict. We cannot guarantee that the order of the fields in `schema.fields` is the same as the order of the keys we get in `signal.init_data.keys()` . 

Anyway, it makes more sense to generate the test data based solely on what's in the signal  (not the schema), since the schema is part of what we're testing. 

In light of that, we should just be able to iterate through everything in the `signal.init_data` dict and use the key for the attribute name and the value as the current_class. this will avoid having to iterate with list indexes and will not depend on our particular implementation of the schema generator.

rgraber (changes requested):
@robrap We probably should document the current need to manually import all  new signal files into the test file to make sure they go through the test, but I'm not really sure where to add that. Thoughts?

robrap:
> @robrap We probably should document the current need to manually import all new signal files into the test file to make sure they go through the test, but I'm not really sure where to add that. Thoughts?

@rgraber: Is this still true? Is there any way around this?

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,62 @@
...
 
-class TestAvro(TestCase):
+class TestAvro(FreezeSignalCacheMixin, TestCase):
     """Tests for end-to-end serialization and deserialization of events"""
+    def test_all_events(self):
+        for signal in OpenEdxPublicSignal.all_events():
+            if signal.event_type not in KNOWN_UNSERIALIZABLE_SIGNALS:
+                test_data = {}
+                serializer = AvroSignalSerializer(signal)
+                for key in list(signal.init_data):
```
rgraber, line 67: nit:
`for key, curr_class in signal.init_data.entries()`

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,62 @@
...
+            dtdict.update({attribute.name: attribute.type(**generate_test_event_data_for_data_type(attribute.type))})
+    return dtdict
 
 
-class TestAvro(TestCase):
+class TestAvro(FreezeSignalCacheMixin, TestCase):
     """Tests for end-to-end serialization and deserialization of events"""
+    def test_all_events(self):
+        for signal in OpenEdxPublicSignal.all_events():
+            if signal.event_type not in KNOWN_UNSERIALIZABLE_SIGNALS:
```
rgraber, line 64: nit: if you want this to be an early return
```
if signal.event in KNOWN_UNSERIALIZABLE_SIGNALS:
    continue
test_data = ...
```

rgraber (approved):
Excellent!

rgraber:
I approved, but would like @robrap 's eyes as well

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,63 @@
...
     serialize_event_data_to_bytes,
 )
+from openedx_events.learning import signals  # pylint: disable=reimported
+from openedx_events.tests.utils import FreezeSignalCacheMixin
+from openedx_events.tooling import OpenEdxPublicSignal
+
+# If a signal is explicitly not for use with the event bus, add it to this list
+KNOWN_UNSERIALIZABLE_SIGNALS = ["org.openedx.learning.discussions.configuration.changed.v1"]
+
+# TODO (EventBus): figure out how to automatically pull new signals in
```
robrap, line 27: [question] Is this still accurate? I'm missing where we are doing this manually. Please remove if it no longer applies.

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,63 @@
...
+def generate_test_event_data_for_data_type(data_type):
+    """
+    Generates test data automatically by filling in dummy data for basic data types (int/bool/str)
+    Recursively breaks down the classes for nested classes
+
+    Arguments:
+        data_type: The type of the data which we are generating data for
+
+    Returns:
+        dtdict: A data dictionary containing dummy data for all attributes of the class
```
robrap, line 39: Just provide the type here. "dtdict" is not really meaningful externally.
```suggestion
        (dict): A data dictionary containing dummy data for all attributes of the class
```

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,63 @@
...
+    """
+    dtdict = {}
```
robrap, line 41: [strong nit] Is "dt" just short for data? Try to avoid abbreviations because they can just be confusing. Just use `data_dict` instead.

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,63 @@
     deserialize_bytes_to_event_data,
     serialize_event_data_to_bytes,
 )
+from openedx_events.learning import signals  # pylint: disable=reimported
+from openedx_events.tests.utils import FreezeSignalCacheMixin
+from openedx_events.tooling import OpenEdxPublicSignal
+
+# If a signal is explicitly not for use with the event bus, add it to this list
```
robrap, line 24: [blocker] 
1. Can you add an additional note like the following
```suggestion
# If a signal is explicitly not for use with the event bus, add it to this list
#   and document why in the event's annotations.
```

2. Also, could you add that documentation to the discussions event you included?

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,63 @@
...
+            test_data = {}
+            serializer = AvroSignalSerializer(signal)
+            for key, curr_class in signal.init_data.items():
+                example_data = generate_test_event_data_for_data_type(curr_class)
+                example_data_processed = curr_class(**example_data)
+                test_data.update({key: example_data_processed})
+            serialized = serializer.to_dict(test_data)
+            deserializer = AvroSignalDeserializer(signal)
+            deserialized = deserializer.from_dict(serialized)
+            print(test_data)
```
robrap, line 75: I assume the print statement was just for debugging? Please remove.

robrap (commented):
Looks great. Minor notes. Thank you.

openedx_events/event_bus/avro/tests/test_avro.py
whuang1202: I was under the impression that for signals that fail the tests, we document it in the known unserializable signals so the tests will always skip these and run tests on only working signals to see if old ones break after new code is put in, or newer signals break, we manually put in the signals onto the KNOWN_UNSERIALIZABLE_SIGNALS list

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,62 @@
...
+# If a signal is explicitly not for use with the event bus, add it to this list
+KNOWN_UNSERIALIZABLE_SIGNALS = ["org.openedx.learning.discussions.configuration.changed.v1"]
+
+# TODO (EventBus): figure out how to automatically pull new signals in
+
+
+def generate_test_event_data_for_data_type(data_type):
+    """
+    Generates test data automatically by filling in dummy data for basic data types (int/bool/str)
+    Recursively breaks down the classes for nested classes
```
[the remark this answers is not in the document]
whuang1202: will do

openedx_events/event_bus/avro/tests/test_avro.py
robrap: I think I am on the same page about your PR comment. But, I still don't understand the purpose of this "TODO" comment in the code. Should it just be removed? If not, can you help me better understand what it means?

robrap (commented):
@whuang1202: 
* Apologies. I did the review and forgot to submit it, so my comments were pending. Here they are. 
* Also, looks like there are some quality test failures.
* Also, a "Draft" PR is usually how we indicate that it isn't ready for review, or that it is getting a really early review. I think you can take this out of draft mode. Thanks.

@rgraber: Could you maybe mark your resolved comments as resolved? I'm not certain that they've all been addressed.

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,58 @@
     deserialize_bytes_to_event_data,
     serialize_event_data_to_bytes,
 )
+from openedx_events.learning import signals  # pylint: disable=reimported
+from openedx_events.tooling import OpenEdxPublicSignal
+
```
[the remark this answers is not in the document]
rgraber: I believe we still need that TODO. We still have to manually import signals files, which is annoying and error prone when someone adds new signals.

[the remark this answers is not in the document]
rgraber: (There was a time when I thought this wasn't true but I was proved wrong)

openedx_events/event_bus/avro/tests/test_avro.py
robrap: @whuang1202: This still has not been addressed.

openedx_events/event_bus/avro/tests/test_avro.py
robrap: @whuang1202: I see you removed this. Does that mean you discussed with Becca and agreed it was no longer relevant? I often find it useful to get PR comment responses to explain the changes.

openedx_events/event_bus/avro/tests/test_avro.py
whuang1202: oh mb I only committed all the files I changed under the avro folder, let me make a commit real quick

openedx_events/event_bus/avro/tests/test_avro.py
whuang1202: I took it out temporarily since it didn't make much sense, I'll put up a clearer todo message for the next commit since we're referring to importing signals

openedx_events/event_bus/avro/tests/test_avro.py
robrap: Thanks. I see the change. Note: from a commit message perspective, you may want to squash this with a "docs:" commit message and the rest in a "test:" commit. However, if this gets too complicated for you, a single commit with a "test:" commit message will be fine.

openedx_events/event_bus/avro/tests/test_avro.py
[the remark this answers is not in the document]
robrap: Let’s discuss. I’m not clear what manual import is being discussed. If the TODO returns:
1. Can it be located with the import to which you are referring?
2. We might need to update the text to make it more clear.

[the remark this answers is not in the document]
robrap: Maybe https://github.com/openedx/openedx-events/pull/68/files#r911418145 is what it is all about?

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -11,10 +17,65 @@
     deserialize_bytes_to_event_data,
     serialize_event_data_to_bytes,
 )
+from openedx_events.learning import signals  # pylint: disable=reimported
```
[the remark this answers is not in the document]
whuang1202: Yes, these are the imports we are currently using to import all the signals despite calling all signals from all_events, I'll update the todo with the comment you suggested.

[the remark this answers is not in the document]
robrap: Looks good. Thanks.

PR #71, opened 2022-06-30
fix: Install pip and pip-tools in upgrade script

Jawayria:


mariajgrimaldi:
Is this PR a duplicate of https://github.com/openedx/openedx-events/pull/70?

Issue #72, opened 2022-07-01
[Data] Create new CMS_COURSE_PUBLISHED signal to represent courses being updated in Studio


rgraber:
- [x] Create a `CMS_COURSE_PUBLISHED` (name open to discussion) OpenEdxPublicSignal in openedx-events with the fields listed below (with appropriate data types)
- [x] Get TNL and tCRIL feedback on the event design and location within openedx-events

Fields:
blocks_url
effort
end
enrollment_start
enrollment_end
id
media: { banner_image: {uri, uri_absolute}, course_image: {uri}, course_video: {uri}, image: {raw, small, large} }
name
number
org
short_description
start
start_display
start_type
pacing
mobile_available
hidden
invitation_only
course_id

Note: CMS_COURSE_PUBLISHED is just a suggestion for a name and can be improved on. We should make sure we are consistent with "CMS" vs "Studio" nomenclature. 

Open Questions:
Where in the openedx-events module structure should this new event live? @ormsbee does tCRIL have a strong opinion on this?

ormsbee:
Also:

1. When is the signal sent?
2. Who are the consumers of the signal?
3. What do those consumers do with the signal data?
4. Do consumers have to make separate requests to get more information?
5. Do different consumers take different subsets of this data? Does it make sense to split it into multiple separate signals?

ormsbee:
For instance, if this is intended to broadcast when Studio users change information that specifically affects catalog listings, it could be renamed to something that more specifically focuses on that scope.

> Where in the openedx-events module structure should this new event live? @ormsbee does tCRIL have a strong opinion on this?

If it's triggered by a Studio action, it probably belongs in a separate `authoring` package, instead of being lumped together with `learning` events that are more LMS-centric.

ormsbee:
I feel like that page is a bit self-contradictory in that in tables it lumps catalog authoring along with course content authoring, but in the top level breakdown images, it puts Discovery/Catalog in an entirely different subdomain.

![openedxddd-diagram2](https://user-images.githubusercontent.com/41625/178323125-002dd58a-7208-4539-ad52-ab52838be2fb.png)

My personal view is that the catalog is its own subdomain, and authoring should be scoped to the edits that happen in Studio (+import/export) to the learning content. I think that in the long term, we'd want a data flow that looks something like:

 ```mermaid
graph LR;
[...]

PR #81, opened 2022-07-21
feat: new COURSE_CATALOG_INFO_CHANGED signal for course_authoring

rgraber:
**Description:** 
Adds a new OpenEdxPublicSignal to describe the changing of the catalog information for a course

**Issue:** https://github.com/openedx/openedx-events/issues/72

**Reviewers:**
- [ ] tag reviewer 
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+    id = attr.ib(type=CourseKey)
+    course_id = attr.ib(type=CourseKey)
+    name = attr.ib(type=str)
+    number = attr.ib(type=str)
+    org = attr.ib(type=str)
+
+    # additional marketing information
+    short_description = attr.ib(type=str, default=None)
+    effort = attr.ib(type=str, default=None)
+    display_name = attr.ib(type=str, factory=str)
```
robrap, line 112: [question] Why is this using factory?

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+        id (str): identifier of the Course object.
+        course_id (str): identifier of the Course object.
+        name (str): course name
+        org (str): course organization identifier
+        number (str): course number
+        short_description (str): one- or two-sentence course description (optional)
+        display_name (str): display name associated with the course.
+        effort (str): estimated level of effort in hours per week (optional). Kept as a str to align with the lms model.
+        schedule_data (CourseScheduleData): scheduling information for the course
+        media (MediaData): associated media for the course (video and images)
```
robrap, line 96:
```suggestion
        media (CourseMediaData): associated media for the course (video and images)
```

openedx_events/content_authoring/signals.py
```diff
@@ -0,0 +1,23 @@
...
+They also must comply with the payload definition specified in
+docs/decisions/0003-events-payload.rst
+"""
+
+from openedx_events.content_authoring.data import CourseCatalogData
+from openedx_events.tooling import OpenEdxPublicSignal
+
+# .. event_type: org.openedx.content_authoring.course.catalog_info.changed.v1
+# .. event_name: COURSE_CATALOG_INFO_CHANGED
+# .. event_description: Fired when a course changes in Studio in a way that is relevant for Catalog consumers.
```
robrap, line 16: 1. [nit]: Questioning capitalization of "Catalog".
2. [questions] Should docs also refer to CMS and not Studio?

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -4,7 +4,11 @@
 
 from opaque_keys.edx.keys import CourseKey
 
-from openedx_events.enterprise import signals  # pylint: disable=unused-import
+# Each new folder with signals must be manually imported in order for the signals to be cached
+#   and used in the unit tests. Using 'disable=reimported' with pylint will work,
+#   because we just use the cached signal list.
+from openedx_events.content_authoring import signals  # pylint: disable=unused-import
+from openedx_events.enterprise import signals  # pylint: disable=reimported
```
robrap, line 11: [suggestion] Maybe add something like the following here, and especially to the `learning` import, to help someone find the relevant comment?
```suggestion
from openedx_events.enterprise import signals  # See note above; pylint: disable=reimported
```

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -17,9 +21,6 @@
     deserialize_bytes_to_event_data,
     serialize_event_data_to_bytes,
 )
-# Each new folder with signals must be manually imported in order for the signals to be cached
-#   and used in the unit tests. Using 'disable=reimported' with pylint will work,
-#   because we just use the cached signal list.
 # TODO (EventBus): Find a way to auto-discover these signals.
```
robrap, line 24: [suggestion] Can we just remove the TODO, and fix when and if we decide it is necessary?
```suggestion
```

robrap (commented):
Nice. Lots of minor questions/comments that hopefully aren't too controversial.

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    Data for a media object (image or video).
+
+    Arguments:
+        uri (str): media object filename
+        absolute_uri (str): fully qualified media object location
+    """
+
+    uri = attr.ib(type=str, default=None)
+    uri_absolute = attr.ib(type=str, default=None)
```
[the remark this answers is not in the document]
rgraber: good catch. just a typo on my end. the real code uses uri_absolute

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Arguments:
+        banner_image (MediaData): top banner image (optional)
+        course_image (MediaData): catalog image (optional)
+        course_video (MediaData): catalog video (optional)
+        image (ImageData): course card image
+    """
+
+    banner_image = attr.ib(type=MediaData, default=None)
+    course_image = attr.ib(type=MediaData, default=None)
```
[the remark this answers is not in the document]
rgraber: I just couldn't think of a way to describe course_image other than "course image," which doesn't seem helpful. Definitely open to suggestions.

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+    course_video = attr.ib(type=MediaData, default=None)
+    image = attr.ib(type=ImageData, default=None)
```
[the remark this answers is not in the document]
rgraber: We could update ImageData to be
`
uri_raw
uri_absolute_raw
uri_small
uri_absolute_small
uri_large
uri_absolute_large
`
it will require a bit of fudging to get all the data in the correct fields but it means we can reuse it.

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Arguments:
+        start (str): course start date
+        end (str): course end date
+        enrollment_start (datetime): start of course enrollment
+        enrollment_end (datetime): end of course enrollment
+        pacing (str): instructor- or self-paced
+    """
+
+    start = attr.ib(type=datetime, default=None)
```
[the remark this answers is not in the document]
rgraber: I'm not 100% sure if start is required but if not it probably should be, so I'm down with removing the default.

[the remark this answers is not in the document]
rgraber: Ah I see I failed to document this properly.

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@attr.s(frozen=True)
+class CourseCatalogData:
+    """
+    Data needed for a course catalog entry.
+
+    Arguments:
+        id (str): identifier of the Course object.
+        course_id (str): identifier of the Course object.
```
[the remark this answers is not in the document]
rgraber: Both of them come from studio and tbh I'm not sure why both fields are included since I think they are identical. I'll just change it to course_key and we can translate it on the consuming side.

openedx_events/content_authoring/data.py
rgraber: copypasta. good catch.

openedx_events/content_authoring/signals.py
rgraber: I used 'Studio' since that is the name of the authoring service, while 'CMS' is the name of the full web service. It's a pretty delicate  distinction. https://github.com/openedx/edx-platform/blob/master/docs/decisions/0013-cms-vs-studio.rst

ormsbee (commented):
Some questions (mostly because I'm not as familiar with this data as I should be).

rgraber:
@ormsbee @robrap 
As I understand it, the problem is that we don't send media across in a consistent format in the LMS API call that we're trying to replace, nor does Discovery save in a consistent format. For `course_video`, we send both relative URI/path and absolute URI/URL. For `banner_image` and `course_image` we just send a path, and for `image` we send over three full URLs for raw, small, and large.

After some discussion, I think our options are
1. Copy the existing fields exactly as they are from the LMS response and just acknowledge that they are weird and inconsistent. We could still fix the field names to be a little clearer.
2. Create an `ImageData` attrs-class with `path`  and `URL` fields. Use this class for `video`, `banner_image`, and `course_image.` Then, create a `SizedImageData` class that is just three `ImageData` attributes, one each for `raw`, `small`, and `large`. Make any bits we can't always populate optional. 
3. Always send just the URL and let clients deal with extracting the path. 

Notes:
For (2), we might end up with a lot of optional fields which isn't great for consumers
For (3), we're not sure if we can get the absolute uri from within Studio

@robrap 's order of preference is (3) then (1). This preference for (3) assumes we are able to get absolute uris from within Studio without too much headache and then strip down to a relative path in Discovery with a similarly small headache. He really doesn't like (2) because of the confusing nature of a bunch of optional fields. 

My order of preference is (1) then (3), though it's not a terribly strong preference and is mostly in service of getting this out quicker. I agree (2) has the potential to be very confusing and doesn't save that much work over (3). If (3) is quick, then I'm all for it, I'm just not sure how long it will take to determine if it's quick.

I think both of us agree we can at least start with 1, and then when implementing the actual event in Studio determine if we could do 3 without much more effort.

openedx_events/content_authoring/data.py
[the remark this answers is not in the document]
robrap: In response to https://github.com/openedx/openedx-events/pull/81#issuecomment-1192730301:

> 1. Copy the existing fields exactly as they are from the LMS response and just acknowledge that they are weird and inconsistent. We could still fix the field names to be a little clearer.

I was wondering if we could start with a slightly edited version of option (1), and skip the current MediaData class. Let's provide the fields directly in the higher level Data class, and always provide exactly one str of either url or path. 

Notes: 
* MultiSizeImageData (or whatever you want to call it), could still be a separate class if you want it.
* If you have both url and path, just supply the url and the path can be calculated by the consumer. (This is a bit of option 3.)
* If you just have path, we'll learn later if it can easily be changed to a url and we can update this definition later.

Of course this all depends on your research of where this data is coming from in the first place. If it doesn't come from Studio editing, it doesn't really belong as part of this event.

rgraber:
As far as I can tell, while it used to be possible to edit media in Studio, it is not possible any more. This would suggest all the various media fields sent over from the lms to discovery are legacy. We could turn on some temporary logging in discovery to see if it ever actually updates course media as a result of RCM to support this hypothesis.

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+        name (str): course name
```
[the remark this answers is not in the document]
rgraber: Good catch. That was a mistake. I don't know where I thought `display_name` came from.

openedx_events/event_bus/avro/tests/test_avro.py
rgraber: Fixed (on the learning import)

CHANGELOG.rst
```diff
@@ -16,6 +16,12 @@ Unreleased
 Changed
 ~~~~~~~
 
+[0.10.1] - 2022-07-21
+---------------------
+Changed
```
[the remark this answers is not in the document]
rgraber: Adopted

openedx_events/content_authoring/data.py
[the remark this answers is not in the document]
rgraber: Removed all the media stuff

openedx_events/content_authoring/data.py
rgraber: fixed

openedx_events/content_authoring/data.py
rgraber: took out the media stuff

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+    uri_absolute = attr.ib(type=str, default=None)
+
+
+@attr.s(frozen=True)
+class ImageData:
+    """
+    Data for a display image.
+
+    Arguments:
+        raw (str): uri of the raw image (required)
```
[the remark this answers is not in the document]
rgraber: removed with the rest of the media stuff

openedx_events/content_authoring/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+Package where events related to the ``content_authoring`` subdomain are implemented.
+
+The ``content_authoring`` subdomain corresponds to {Architecture Subdomain} defined in
+the OEP-41.
```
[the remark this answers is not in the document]
rgraber: added

openedx_events/content_authoring/data.py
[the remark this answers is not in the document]
rgraber: Resolved by dropping media stuff

openedx_events/content_authoring/data.py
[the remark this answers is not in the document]
rgraber: Made required.

openedx_events/content_authoring/data.py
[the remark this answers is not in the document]
rgraber: resolved by just using course_key

openedx_events/content_authoring/signals.py
rgraber: marking as resolved since i think i answered the question

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@attr.s(frozen=True)
+class CourseScheduleData:
+    """
+    Data describing course scheduling.
+
+    Arguments:
+        start (str): course start date
+        end (str): course end date
```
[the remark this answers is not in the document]
rgraber: yes, fixed

rgraber:
For posterity, the decision was to drop all the media fields since they are unused by course-discovery. See the deprecation ticket at https://github.com/openedx/public-engineering/issues/160 .

openedx_events/content_authoring/data.py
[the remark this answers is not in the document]
rgraber: Resolved by dropping media

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,117 @@
...
+        enrollment_start (datetime): start of course enrollment
+        enrollment_end (datetime): end of course enrollment
+        pacing (str): instructor- or self-paced
```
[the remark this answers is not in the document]
rgraber: updated to use the real possible values

openedx_events/event_bus/avro/tests/test_avro.py
rgraber: added to the learning import

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # basic identifiers
+    course_key = attr.ib(type=CourseKey)
+    name = attr.ib(type=str)
+    number = attr.ib(type=str)
+    org = attr.ib(type=str)
+
+    # additional marketing information
+    short_description = attr.ib(type=str, default=None)
+    effort = attr.ib(type=str, default=None)
```
ormsbee, line 61: Is there a standard unit of "effort"? Or is this a freeform description left up to the course team?

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,64 @@
...
+    schedule_data = attr.ib(type=CourseScheduleData, default=None)
```
ormsbee, line 62: Is it legal/well formed to have `None` here for `schedule_data`?

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,64 @@
...
+    Arguments:
+        start (datetime): course start date
+        end (datetime): course end date
+        pacing (str): 'instructor' or 'self'
+        enrollment_start (datetime): start of course enrollment (optional)
+        enrollment_end (datetime): end of course enrollment (optional)
+    """
+
+    start = attr.ib(type=datetime)
+    end = attr.ib(type=datetime)
```
ormsbee, line 30: Is `end` `None` when courses are self-paced?

ormsbee: If so, you may want to make the type `Optional[datetime]` to indicate that `None` is a legal value.

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,64 @@
...
+
+@attr.s(frozen=True)
+class CourseScheduleData:
+    """
+    Data describing course scheduling.
+
+    Arguments:
+        start (datetime): course start date
+        end (datetime): course end date
+        pacing (str): 'instructor' or 'self'
```
ormsbee, line 24: Do you think it's worth having this be an Enum? We have an example of that kinda usage in edx-platform: https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/content/learning_sequences/data.py#L35-L38

openedx_events/content_authoring/data.py
rgraber: Sometimes. It's unfortunately rather variable.

robrap: Start and end are documented as "optional", so should probably use `default=None`?

@ormsbee: See this ADR and the final section around punting on implementing `Optional`: https://github.com/openedx/openedx-events/blob/main/docs/decisions/0007-optional-fields.rst#deferredrejected-decisions Feel free to disagree, but that is where this was discussed.

openedx_events/content_authoring/data.py
robrap: Also note that docstring does not say this is optional.

openedx_events/event_bus/avro/tests/test_avro.py
robrap, line 11: Nit: I'd add this to this line as well, but we'll see how strongly you disagree. ;)
```suggestion
from openedx_events.enterprise import signals  # See note above; pylint: disable=reimported
```

openedx_events/content_authoring/data.py
robrap: Not sure about de/serialization of enums, and how it would effect event evolution. In other words, can we simply add enum options under different evolution settings without complaint. 
@ormsbee: Can we punt on this question until later?
@rgraber: Or, do we have a quick way to answer these types of questions?

robrap (commented):
I'm basically all set, other than resolving Dave's comments. Thank you.

openedx_events/content_authoring/data.py
rgraber: Right now the serializer cannot handle enum fields.

openedx_events/content_authoring/data.py
rgraber: It's supposed to be hours per week, but the field in the lms is just freeform text

openedx_events/event_bus/avro/tests/test_avro.py
rgraber: I disagree strongly enough to not want to do it but not strongly enough to spend more than this comment arguing about it. I think it's unnecessary since it's just one line away.

openedx_events/content_authoring/data.py
rgraber: forgot about that. I'll make the end date optional. The start date should be kept mandatory since I'm fairly certain it's only nullable on legacy courses.

openedx_events/content_authoring/data.py
rgraber: Probably not a good idea to have it as None. I'll make it mandatory.

openedx_events/content_authoring/data.py
ormsbee: I hadn't seen the ADR linked (thank you!), or maybe I saw it but forgot. I had personally been differentiating between "these are legal permutations of the data" (`type=Optional[datetime]`) vs. "we offer these convenient defaults for backwards compatibility or to make message creation easier". It looks like those are purposely conflated in ADR-0007. I have to think about it more, but my only longer term concern about that would be whether it complicates mypy style checking of type information. 

I'm also a bit confused about the wording of:

> We are also deferring any work around making an explicit `Optional` keyword for attrs fields

Is this referring to using `Optional` from the Python's `typing` module, or is it implying that there would have to create something new for some reason...?

Anyhow, these seem reversible decisions, and I definitely don't want to hold up this PR in order to re-hash an ADR. I withdraw my suggested change.

openedx_events/content_authoring/data.py
ormsbee: Got it. Makes sense.

openedx_events/content_authoring/data.py
ormsbee: We'd probably serialize to string anyhow. It's mostly about making sure people don't create events with typos that get lost. And yeah, I think we can punt this for now.

ormsbee: FWIW, in the longer term, I think as long as we strictly add to the enum and don't change old enum values, we can evolve it in a way that's consistent with the evolution of our other data structures.

openedx_events/event_bus/avro/tests/test_avro.py
robrap: No worries. Skip it.

openedx_events/content_authoring/data.py
robrap: I meant can we make Avro's evolution rules happy, even if it makes sense to us to be able to update. I'm not sure how enums work in Avro, but looks like we are punting for now.

openedx_events/content_authoring/data.py
rgraber: Sounds like this conversation is resolved? Marking it as such.

openedx_events/content_authoring/data.py
rgraber: Made mandatory

openedx_events/content_authoring/data.py
rgraber: Punting.

openedx_events/content_authoring/data.py
robrap: Re: ADR and `Optional`:
I'm guessing we were not clear other than to want to put this off. I'm guessing the ADR could more clearly state that `Optional` could be from Python's `typing` module, but it is still unclear how this all would work.

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,64 @@
...
+
+@attr.s(frozen=True)
+class CourseScheduleData:
+    """
+    Data describing course scheduling.
+
+    Arguments:
+        start (datetime): course start date
+        pacing (str): 'instructor' or 'self'
+        end (datetime): course end date (optional)
```
timmc-edx, line 24: I noticed a change to move `pacing` to be between `start` and `end`, which I think makes it a little harder to read. Is there something I'm missing?

robrap (approved):

openedx_events/content_authoring/data.py
robrap: Note: I was going to write a similar comment, and then realized it was probably grouping required vs optional (which wasn't done on the other data class). I decided to not comment, but keeping start/end together, which share the same type would also make sense to me.

rgraber: non-optional arguments before optional ones

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,64 @@
...
+@attr.s(frozen=True)
+class CourseCatalogData:
+    """
+    Data needed for a course catalog entry.
+
+    Arguments:
+        course_key (CourseKey): identifier of the Course object.
+        name (str): course name
+        org (str): course organization identifier
+        number (str): course number
```
timmc-edx, line 45: [nit] org/number are in a different order here than in the attributes.

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,64 @@
...
+        schedule_data (CourseScheduleData): scheduling information for the course
+        short_description (str): one- or two-sentence course description (optional)
+        effort (str): estimated level of effort in hours per week (optional). Kept as a str to align with the lms model.
+        hidden (bool): whether the course is hidden from search (optional)
```
timmc-edx, line 49: [nit] `hidden` uses `(optional)` on a bool with a default, but `invitation_only` does not.
```suggestion
        hidden (bool): whether the course is hidden from search (optional)
```

openedx_events/content_authoring/data.py
```diff
@@ -0,0 +1,64 @@
...
+        invitation_only (bool): whether the course requires an invitation to enroll
+    """
+
+    # basic identifiers
+    course_key = attr.ib(type=CourseKey)
```
timmc-edx, line 54: Ah neat, I guess this works because CourseKey has string de/serializer methods?

robrap: We have a custom serializer: https://github.com/openedx/openedx-events/blob/main/openedx_events/event_bus/avro/custom_serializers.py#L34-L50

rgraber:
The eventual ADR around this event design ended up here: https://github.com/openedx/openedx-events/pull/114/files

PR #83, opened 2022-07-28
test: Add unit test locking in KeyError when signal not found

timmc-edx:
I wanted to start relying on the KeyError not-found behavior in another library and realized it wasn't actually documented or tested.


mariajgrimaldi (approved):
This looks good! Thank you for noticing

PR #85, opened 2022-07-28
feat: Configurable loading of event bus producer

timmc-edx:
This defines an API for IDAs to use when producing events that allows the
choice of Event Bus implementation to be configured rather than written in
code. Implementations can document the appropriate setting for deployers
to use.

Caching the producer instance means that we automatically get a
long-lived instance without the implementation itself having to
perform caching itself.

This is part of https://github.com/openedx/openedx-events/issues/87

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed


openedx_events/event_bus/__init__.py
```diff
@@ -1 +1,83 @@
...
+def _load_event_bus() -> EventBus:
+    """
+    Load the currently configured event bus implementation.
+
+    If configuration is invalid, will return a NoEventBus instance that does nothing.
+    """
+    try:
+        eb = settings.EVENT_BUS
+        loader = import_string(eb['loader'])
+        impl = loader(eb)
```
timmc-edx, line 72: [author note] I'm imagining that the `EVENT_BUS` dictionary setting might carry the additional implementation-specific settings that the loader function will then want to read. So we might have:

```
EVENT_BUS = {
  "loader": "edx_event_bus_kafka.create",
  "env_prefix": "prod",
  "kafka": {
    "schema_registry": "...",
    "bootstrap_servers": "..."
  }
}
```

openedx_events/event_bus/__init__.py
```diff
@@ -1 +1,83 @@
...
+            return impl
+        else:
+            warnings.warn(f"Loader {loader} a {type(impl)}, not an EventBus", UserWarning)
+            return NoEventBus()
+    except BaseException as e:
+        warnings.warn(f"Failed to load event bus implementation: {e!r}", UserWarning)
+        return NoEventBus()
+
+# The configured event bus implementation, or a stub instance that does nothing.
+EVENT_BUS = _load_event_bus()
```
robrap, line 83: We had punted on caching settings to ensure we supported remote config changes. If we feel we can do this quickly, you might want to look for patterns in DRF settings, where they use the settings_changed signal from Django to reload: https://github.com/encode/django-rest-framework/blob/master/rest_framework/settings.py.

openedx_events/event_bus/__init__.py
robrap: Generally I think this is ok, but I find the naming a bit confusing.
1. Instead of "loader" setting, could it just be "event_bus_class" or something? What is `create`, and could we just use provide a path to the class instead?
2. Do we need to pass the settings to the "loader"? Could we use something like the DRF settings, as noted in the other comment?

openedx_events/event_bus/__init__.py
```diff
@@ -1 +1,83 @@
...
+from django.utils.module_loading import import_string
+
+
+class EventBus(ABC):
+    """
+    Parent class for event bus implementations.
+    """
+
+    @abstractmethod
+    def send(
```
robrap, line 16: Just confirming that this replaces the previously documented `send_to_event_bus`, since this would probably result in code like `event_bus.send(...)`?
Related, do we like using the term `send` rather than `produce`, to provide overlap with the signals, since this takes a signal? Or would something like `produce_from_signal` be more clear?

openedx_events/event_bus/__init__.py
```diff
@@ -1 +1,83 @@
...
+            *, signal: OpenEdxPublicSignal, topic: str, event_key_field: str, event_data: dict,
+            sync: bool = False,
+    ) -> None:
+        """
+        Send a signal event to the event bus under the specified topic.
+
+        Arguments:
+            signal: The original OpenEdxPublicSignal the event was sent to
+            topic: The event bus topic for the event
+            event_key_field: Path to the event data field to use as the event key (period-delimited
```
robrap, line 26: Is this already implemented, or being introduced here? Is there an advantage to pulling this out based on some path, vs just having the caller pull it out and send the actual object they want to send?

openedx_events/event_bus/__init__.py
```diff
@@ -1 +1,83 @@
...
+              string naming the dictionary keys to descend)
+            event_data: The event data (kwargs) sent to the signal
+            sync: Whether to wait indefinitely for event to be received by the message bus (probably
+              only want to use this for testing)
```
robrap, line 30: Is this something we will be using now? What does this translate to? An immediate call to `flush()`? Do we need to expose `flush`? Do we want to separate Producer and Consumer code in case either gets more complex and there is confusing overlap?

robrap (commented):
Generally right direction. Just digging into details.

openedx_events/event_bus/__init__.py
timmc-edx: No strong opinion! But yeah, it seemed redundant to use the existing name.

openedx_events/event_bus/__init__.py
timmc-edx: Already implemented in event-bus-kafka.

It's entirely possible that in many/most cases, no IDA will actually be calling this directly, and it will just be called by openedx-events based on settings. (openedx-events reads configuration, learns what signals to add receivers to, and adds receivers that convey these settings to event_bus.send.)

timmc-edx: I think in a previous PR or meeting we discussed having this be a function instead. Maybe we defer that discussion until we need something different?

openedx_events/event_bus/__init__.py
timmc-edx: For Kafka, a call to flush. But... maybe there should instead be a different method on `EventBus` called `shutdown` or something, which can be used for testing (and later be called by some Django shutdown hook.)

At first thought I'd prefer having this be the single entry point for both consumer and producer so we don't have to load multiple things. But hmm... for Kafka it's all the same; once you have the connection settings, I think everything else is optional. I suppose it's possible that with some other implementation you might have a well-configured producer and an ill-configured consumer, and if you really just want the producer side then you wouldn't care! So it might be worth splitting after all.

openedx_events/event_bus/__init__.py
timmc-edx: That exactly the sort of thing I'd want, except it looks like it's only called during tests. Do we arrange for it to be signaled by remote-config as well?

openedx_events/event_bus/__init__.py
timmc-edx: 1. We could just use the class constructor, yeah. So `edx_event_bus_kafka.EventBusKafka`.
2. I suppose we don't need to pass the settings. The implementation could look them up in `django.conf.settings` itself.

openedx_events/event_bus/__init__.py
timmc-edx: In event-bus-kafka's implementing class I'd love to be able to do something like this:

```
def __init__(...):
    self.producer = create_producer()
    ...and same for consumer
    setting_changed.connect(lambda: self.reload_config())

def reload_config(self):
    old_producer = self.producer
    self.producer = create_producer()
    threading.Thread(target=lambda: old_producer.flush()).start()
    ...and same for consumer
```
...but that assumes we can actually get such a signal.

robrap: This is the code I was referring to with the signal: https://github.com/encode/django-rest-framework/blob/a1b35bb44b7c9251c8b7bc995aa6598044f1d3ef/rest_framework/settings.py#L256. Are you saying something different?

timmc-edx: I'm saying that Django does not send that signal except during tests, so we can't rely on it in a running service.

openedx_events/event_bus/__init__.py
timmc-edx: 48e3918 splits it into a producer class (providing `send`) and a consumer class (providing `consume_indefinitely`). It also removes `sync`; to provide that functionality, we could later add a synchronous `shutdown()` method, which is really what it's for (besides testing).

openedx_events/event_bus/__init__.py
timmc-edx: In https://github.com/openedx/openedx-events/issues/86 it has been suggested that OpenEdxPublicSignal could carry the topic and key-field... although I suppose we could later make it possible to override both topic and key (actual key, not key field).

openedx_events/event_bus/__init__.py
timmc-edx: 48e3918 introduces `get_consumer()` and `get_producer()`, both memoized, rather than a load-time global. (Just to make testing easier! But maybe that's silly.)

openedx_events/event_bus/__init__.py
timmc-edx: In 48e3918 it is now `EVENT_BUS_PRODUCER` and `EVENT_BUS_CONSUMER` rather than `loader`. No more passing settings. Loader can be either a class (acting as a constructor) or a function.

mphilbrick211:
Hi @timmc-edx! Just checking to see if you are planning to proceed with this PR?

timmc-edx:
Yep! This work was deprioritized and has bit-rotted a bit, but I'll be revisiting it.

robrap:
@timmc-edx: We have https://github.com/openedx/openedx-events/issues/87 which mentions wip PRs. You are welcome to close the PRs and verify the issue would help someone find them when and if the work is picked up. That said, do whatever feels right.

timmc-edx:
I pushed up something of a rewrite of the loader code. It maybe that some of it should get pushed into an `internal` module, but I think this will do what we need and allow us to remove Kafka references from IDAs.

openedx_events/event_bus/tests/test_loader.py
```diff
@@ -0,0 +1,118 @@
...
+
+        with assert_warnings([]):
+            # Nothing thrown, no warnings.
+            assert producer.send(
+                signal=SESSION_LOGIN_COMPLETED, topic='user-logins',
+                event_key_field='user.id', event_data={},
+            ) is None
+
+
+def TestConsumer(TestCase):
```
timmc-edx, line 105: Wow, the coverage check was important here -- this was supposed to be a class. :grimacing:

openedx_events/event_bus/tests/test_loader.py
```diff
@@ -0,0 +1,118 @@
...
+    def test_unconfigured(self):
+        with assert_warnings(["Event Bus setting DOES_NOT_EXIST is missing; component will be inactive."]):
+            loaded = _try_load(
+                setting_name="DOES_NOT_EXIST", args=(1), kwargs={'2': 3},
+                expected_class=dict, default={'def': 'ault'},
+            )
+        assert loaded == {'def': 'ault'}
+
+    @override_settings(EB_LOAD_PATH='builtins.dict')
+    def test_success(self):
```
timmc-edx, line 34: Coverage test also found this duplicate name. :sweat_smile:

timmc-edx:
This is ready for review again, and is accompanied by https://github.com/openedx/event-bus-kafka/pull/14

I'm a little unsure about what the consumer API should look like, and it might be better to move that to a new PR later. It might be better to just focus on the producer part right now.

robrap:
@timmc-edx: I think we decided we would close/defer this? We could create a separate issue, or just use this PR as the issue, if you are ok with long-standing PRs. We should add backlog label and move to Future milestone, if you agree with status. Thanks!

timmc-edx:
This one is just for loading of the implementation -- and for now, scoped down to just the producing side. I'm moving the consumer side into a separate branch for later consideration.

rgraber (commented):
I think this is worth a small how-to with examples of the kinds of things you might put in the EVENT_BUS_PRODUCER setting. The how-to could basically copy the setting description but I think a how-to would be a more accessible place for it.

openedx_events/event_bus/__init__.py
```diff
@@ -1 +1,117 @@
...
+    returned instead.
+
+    Arguments:
+        setting_name: Name of a Django setting containing a dotted module path, indicating a callable
+        expected_class: The callable must produce an instance of this class object (or a subclass)
+        default: Object to return if any part of the lookup or loading fails
+    """
+    constructor_path = getattr(settings, setting_name, None)
+    if constructor_path is None:
+        warnings.warn(f"Event Bus setting {setting_name} is missing; component will be inactive.", UserWarning)
```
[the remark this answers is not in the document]
timmc-edx: Ah thanks, yeah, for some reason I was remembering it as a required argument.

()It can be passed as either arg or kwarg here as `warn` doesn't specify positional-only or keyword-only args.)

openedx_events/event_bus/__init__.py
```diff
@@ -1 +1,117 @@
...
+        return default
+
+    try:
+        constructor = import_string(constructor_path)
+        instance = constructor()
+        if isinstance(instance, expected_class):
+            return instance
+        else:
+            warnings.warn(
+                f"{constructor_path} from {setting_name} returned unexpected type {type(instance)}",
```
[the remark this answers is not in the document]
timmc-edx: Good idea, updated.

timmc-edx:
@rgraber Hmm. I feel like it wouldn't be that helpful to give an example of the setting without backing code. The producer setting already gives an example of `some.module.path.EventBusImplementation` but of course there's no `EventBusImplementation` to look at.  Perhaps the documentation could instead link to event-bus-kafka's documentation as an example? It's all a bit chicken-and-egg, but I could come back and make a new PR here once https://github.com/openedx/event-bus-kafka/pull/14/files#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168f is merged.

openedx_events/event_bus/__init__.py
robrap: Let's defer to #86. Thanks.

rgraber:
> @rgraber Hmm. I feel like it wouldn't be that helpful to give an example of the setting without backing code. The producer setting already gives an example of `some.module.path.EventBusImplementation` but of course there's no `EventBusImplementation` to look at. Perhaps the documentation could instead link to event-bus-kafka's documentation as an example? It's all a bit chicken-and-egg, but I could come back and make a new PR here once https://github.com/openedx/event-bus-kafka/pull/14/files#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168f is merged.

I think coming back once that PR is merged makes sense

rgraber (approved):
LGTM

timmc-edx:
OK, I'll add a ticky box for that on the issue.

PR #95, opened 2022-08-16
feat!: Removed `org` and `number` fields from `CourseCatalogData`

timmc-edx:
These are redundant with course key and can be added back in later if we
actually need them.

This is a breaking change but it should only affect event-bus code that
hasn't yet been merged.

See discussion here:
https://github.com/openedx/openedx-events/issues/72#issuecomment-1212177933

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,8 +13,13 @@ Change Log
 
 Unreleased
 ----------
+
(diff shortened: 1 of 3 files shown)


rgraber (approved):
LGTM

PR #104, opened 2022-09-07
build: update make upgrade command

UsamaSadiq:
### Description
- Updated the `make upgrade` target to test installing the `pip.txt` and `pip-tools.txt` to verify if the updated `pip && pip-tools` versions are compatible with each other.
- Added `common_constraints.txt` in the requirements folder to follow global edX constraints.

### TODO
- Removing `django<3` constraint will update the Django to next major version which would be breaking/major change so that'll be done in a separate PR.


mariajgrimaldi (approved):
This looks good! Thanks again. And good call updating Django in another PR! Let me know if I can help you with it.

PR #107, opened 2022-09-08
Upgrade from Django 2.0 to Django 3.0

UsamaSadiq:
### Description
- Subsequent PR after https://github.com/openedx/openedx-events/pull/104 to Upgrade the package from `Django 2.0` to `Django 3.0`.


mariajgrimaldi:
Hello there! I'm tagging you here, so you're aware of this change @timmc-edx @robrap @felipemontoya. We're currently testing against an old Django version, and this PR updates it according to the constraints of edx-platform@master.

robrap:
Ok. Thank you.

tox.ini
```diff
@@ -40,7 +40,7 @@ norecursedirs = .* docs requirements site-packages
 deps =
     -r{toxinidir}/requirements/test.txt
 commands =
-    pytest {posargs}
```
mariajgrimaldi, line 43: what about this change? :thinking:

UsamaSadiq: Explicitly mentioning the target directories is a standard practice being followed across edX/OpenedX repos so I updated it accordingly.

mariajgrimaldi: Good! Thanks!

mariajgrimaldi (approved):

PR #105, opened 2022-09-07
Fix github url strings (org edx -> openedx)

sarina:
## This PR was autogenerated

This pr replaces the old GitHub organization, github.com/edx, with the new GitHub organization, github.com/openedx.

Tagging @openedx/tcril-engineering for review, but others are welcome to provide review.

Ref: https://github.com/openedx/tcril-engineering/issues/42

PR #112, opened 2022-09-12
Update github paths

sarina:
Hi @mariajgrimaldi - I closed #105 in favor of this because the rebase was messy. Could you take another look?


sarina:
For keeping track - https://github.com/openedx/tcril-engineering/issues/42 as parent issue

PR #114, opened 2022-09-13
docs: ADR for course-catalog-info-changed event design

rgraber:
**Description:** Adds an ADR explaining the rationale for designing the course-catalog-info-changed event in such a way as to only work when Publisher is enabled.

Since it's just a docs update I didn't update the version

**Reviewers:**
- [ ] tag reviewer 
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/docs/decisions/0009-course-catalog-info-changed-design.rst
+++ b/docs/decisions/0009-course-catalog-info-changed-design.rst
@@ -1,0 +1,37 @@
+8. Event design of the COURSE_CATALOG_INFO_CHANGED signal
+=========================================================
+
+Status
+------
+
+Provisional
+
+Context
+-------
+
+The COURSE_CATALOG_INFO_CHANGED signal is used by Studio to convey that information relevant to the course catalog has changed. It was created with an eye towards eventually replacing the refresh-course-metadata batch job, which syncs all data between edx-platform and course-discovery.
+
+Refresh-course-metadata functions differently depending on whether or not the system has Publisher enabled. Specifically, after requesting information from the /courses endpoint in Studio, Discovery will ignore certain fields if Publisher is enabled. These fields are mostly part of the `media` attribute of the API response. They are sent in a variety of ways (absolute urls, paths, sometimes divided by size, etc.)
+
+
+Decision
+--------
+
+The COURSE_CATALOG_INFO_CHANGED will only contain the information necessary to work in a Publisher-enabled environment. In particular, this means it will not contain the `media` field usually present in the Studio /courses API endpoint.
+
+In Discovery, if Publisher is not enabled, the consumer will simply ignore the event and not try to update anything.
+
+Rationale
+---------
+
+The way we update media information in refresh-course-metadata is quite haphazard, with no real standard way of sending over the information or of storing it on the Discovery end. Replicating these structures in the COURSE_CATALOG_INFO_CHANGED signal would make the data definition confusing. Moreover, edx.org has Publisher enabled and would thus ignore these fields anyway.
+
+Deferred work
+-------------
+It is possible in the future we will want to use this event to sync data between Studio and Discovery in systems that do not include Publisher. At that time, we can work on cleaning up the media data structures for cleaner event definitions.
+
+Github discussion
+-----------------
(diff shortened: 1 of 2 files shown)


docs/decisions/0009-course-catalog-info-changed-design.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Refresh-course-metadata functions differently depending on whether or not the system has Publisher enabled. Specifically, after requesting information from the /courses endpoint in Studio, Discovery will ignore certain fields if Publisher is enabled. These fields are mostly part of the `media` attribute of the API response. They are sent in a variety of ways (absolute urls, paths, sometimes divided by size, etc.)
+
+
+Decision
+--------
+
+The COURSE_CATALOG_INFO_CHANGED will only contain the information necessary to work in a Publisher-enabled environment. In particular, this means it will not contain the `media` field usually present in the Studio /courses API endpoint.
+
+In Discovery, if Publisher is not enabled, the consumer will simply ignore the event and not try to update anything.
```
robrap, line 22: I may be missing something, but I don't think we should do this. It feels like we should error instead (for now). It feels like the biggest problem would be a misconfiguration where you are in a publisher environment, but these events were being ignored. Maybe we'd just want to improve the error messaging? Are there cases where it is important for this consumer to be running and not doing these updates?

docs/decisions/0009-course-catalog-info-changed-design.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Rationale
+---------
+
+The way we update media information in refresh-course-metadata is quite haphazard, with no real standard way of sending over the information or of storing it on the Discovery end. Replicating these structures in the COURSE_CATALOG_INFO_CHANGED signal would make the data definition confusing. Moreover, edx.org has Publisher enabled and would thus ignore these fields anyway.
```
robrap, line 27: Rather than mentioning edx.org, is it accurate to say we believe that Refresh-course-metadata is only used by organizations with Publisher enabled. (Maybe we can circulate this assumption and ADR to double-check?)

robrap (commented):
Nice. Thanks.

docs/decisions/0009-course-catalog-info-changed-design.rst
rgraber: We would definitely want to at least improve the error messaging so we don't flood the logs with a bunch of key errors. I think it's more worth a warning than an error, which would basically be ignore-but-log-better.

docs/decisions/0009-course-catalog-info-changed-design.rst
rgraber: I actually don't think that's true. Among other things, RCM is more important when there is no publisher since people are more likely to be editing course dates and media in studio.

docs/decisions/0009-course-catalog-info-changed-design.rst
robrap: That sounds fine. I just didn't want to silently ignore a likely problematic state.

docs/decisions/0009-course-catalog-info-changed-design.rst
robrap: How about something like?:
```suggestion
The way we update media information in refresh-course-metadata is quite haphazard, with no real standard way of sending over the information or of storing it on the Discovery end. Replicating these structures in the COURSE_CATALOG_INFO_CHANGED signal would make the data definition confusing.

Until these fields are needed by a consumer in a non-Publisher environment (if ever), it makes sense to defer trying to find a good solution to this problem.
```

docs/decisions/0009-course-catalog-info-changed-design.rst
rgraber: fixed

robrap (approved):
Could you either post a link to this PR from the two relevant github discussions, or post a link to the ADR after it merges. Either way the ADR can be updated if need be.

PR #116, opened 2022-09-16
fix: broken upgrade script to use proper syntax

sarina:


mariajgrimaldi (approved):
thank you!

.github/workflows/upgrade-python-requirements.yml
```diff
@@ -2,20 +2,19 @@ name: Upgrade Python Requirements
 
 on:
   schedule:
-    # will start the job at 13:00 UTC every Friday
-    - cron: "0 13 * * 5"
+    - cron: "0 0 * * 1"
```
sarina, line 5: @mariajgrimaldi I realize I didn't put back your default here. You may want to I have no idea

mariajgrimaldi: I noticed, but that works too so it's okay.

PR #119, opened 2022-09-19
fix: github workflow -branch to actual default branch

sarina:
See https://github.com/openedx/docs.openedx.org/pull/135 for more details


sarina:
apparently @kdmccormick is scooping me on these.

dupe of https://github.com/openedx/openedx-events/pull/118

PR #123, opened 2022-09-22
docs: Bump version to 1.x to acknowledge that this is in use in production

timmc-edx:
According to semver, 0.x versions may have breaking changes at any time, and there's no way to communicate them via the version number until the version is at least 1.0.0. Since this is in active use now, we should bump to 1.0.0.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,11 @@ Change Log
 
 Unreleased
 ----------
+
+[1.0.0] - 2022-09-27
+--------------------
 * **Breaking Change**: Updated from ``Django 2.0`` to ``Django 3.0``.
+* Bump version to 1.x to acknowledge that this is in use in production
 
 [0.14.0] - 2022-09-21
 ---------------------
--- a/openedx_events/__init__.py
+++ b/openedx_events/__init__.py
@@ -5,4 +5,4 @@ These definitions are part of the Hooks Extension Framework, see OEP-50 for
 more information about the project.
 """
 
-__version__ = "0.14.0"
+__version__ = "1.0.0"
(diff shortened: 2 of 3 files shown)


felipemontoya (approved):
:+1: I'm in full support of this. This library is at a stage where nothing should be changing with ample notice. Much less breaking changes.

mariajgrimaldi:
can you also update the setup.cfg file? please! :+1:

timmc-edx:
Oh! I didn't know the version was being recorded there as well. Is there some additional tooling we're supposed to be using?

mariajgrimaldi:
@timmc-edx: we have `bumpversion` configured for both openedx-events & openedx-filters. But it makes more sense to me to come to an agreement and document the process of releasing a new version instead of using what's configured just because

felipemontoya:
Would it make sense to have an action that we could run manually to run bumpversion from the github UI?

timmc-edx:
I'm not sure I really understand the usecase of bumpversion -- is it when there is an unavoidable duplication of the version string to multiple files? Most projects in openedx seem to set the version in the `<root_module>/__init__.py` and then read it from `setup.py`.

I think the most valuable thing would actually be a CI check that validates that all the copies of the version string are in sync, since it's easier (in my opinion) to edit text files than to poke around in the GitHub UI. :-P But I don't see a way to do that with bumpversion.

timmc-edx:
(in any case, probably worth a different PR to address that)

mariajgrimaldi:
@timmc-edx agreed! Thanks again for the version bump

PR #131, opened 2022-10-18
feat!: remove effort field from CourseCatalogData

rgraber:
**Description:**
Removes "effort" field in CourseCatalogData. This field is not actually editable in Studio in setups where Publisher is enabled. Also updates the ADR around COURSE_CATALOG_INFO_CHANGED event design to clarify the fields not being copied from the courses API endpoint.

Note that this is in fact a breaking change and will require some work in edx-platform (and possibly discovery) to bring it in. 

Issue: https://github.com/openedx/openedx-events/issues/133

**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 ----------
 
+[3.0.0] - 2022-10-19
+--------------------
+* **Breaking change**: Removed (optional) field ``effort`` from ``CourseCatalogData.`` Nothing should be relying on this field as it is not used by Course Discovery in Publisher-enabled setups.
+
 [2.0.0] - 2022-10-18
 --------------------
 * **Breaking change**: Removed signal ``SUBSCRIPTION_LICENSE_MODIFIED`` and corresponding data class ``SubscriptionLicenseData``. This should only affect experimental event-bus code (which should also have been deleted by now).
--- a/docs/decisions/0009-course-catalog-info-changed-design.rst
+++ b/docs/decisions/0009-course-catalog-info-changed-design.rst
@@ -11,19 +11,23 @@ Context
 
 The COURSE_CATALOG_INFO_CHANGED signal is used by Studio to convey that information relevant to the course catalog has changed. It was created with an eye towards eventually replacing the refresh-course-metadata batch job, which syncs all data between edx-platform and course-discovery.
 
-Refresh-course-metadata functions differently depending on whether or not the system has Publisher enabled. Specifically, after requesting information from the ``/courses`` endpoint in Studio, Discovery will ignore certain fields if Publisher is enabled. These fields are mostly part of the ``media`` attribute of the API response. They are sent in a variety of ways (absolute urls, paths, sometimes divided by size, etc.)
+Refresh-course-metadata functions differently depending on whether or not the system has Publisher enabled. Specifically, after requesting information from the ``/courses`` endpoint in Studio, Discovery will ignore certain fields if Publisher is enabled. Most, though not all, of these fields are sent as part of the ``media`` attribute of the API response. These media fields are sent in a variety of ways (absolute urls, paths, sometimes divided by size, etc.)
 
 Decision
 --------
 
-The COURSE_CATALOG_INFO_CHANGED will only contain the information necessary to work in a Publisher-enabled environment. In particular, this means it will not contain the ``media`` field usually present in the Studio ``/courses`` API endpoint.
+The COURSE_CATALOG_INFO_CHANGED will only contain the information necessary to work in a Publisher-enabled environment. In particular, this means it will not include some fields usually present in the Studio ``/courses`` API endpoint, for example:
+- ``media``
+- ``short_description``
+- ``mobile_available``
+- ``effort``
 
 In Discovery, if Publisher is not enabled, the consumer will log a warning and not try to update anything.
 
 Rationale
 ---------
 
-The way we update media information in refresh-course-metadata is quite haphazard, with no real standard way of sending over the information or of storing it on the Discovery end. Replicating these structures in the COURSE_CATALOG_INFO_CHANGED signal would make the data definition confusing.
+The way we update media information in refresh-course-metadata is quite haphazard, with no real standard way of sending over the information or of storing it on the Discovery end. Replicating these structures in the COURSE_CATALOG_INFO_CHANGED signal would make the data definition confusing. In addition, it is very difficult to test these other fields without a real non-Publisher environment that runs refresh-course-metadata.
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 ----------
 
+[2.1.0] - 2022-10-19
+--------------------
+* **Breaking change**: Removed field ``effort`` from ``CourseCatalogData``
```
robrap, line 19: _If this is a breaking change:_
1. 3.0.0 should be used for breaking changes.
2. You could add more notes to help people understand that this breaking change probably doesn't affect anyone. See 2.0.0 change note.
3. Should use "feat!:" to designate breaking change.
4. Maybe add "optional", as in: "Removed optional field...".

Note: from the standpoint of anyone that was sending effort and now it is no longer appearing, and they still want it, this would be a breaking change. From the perspective of breaking code or services, in theory, removing an optional field is not a breaking change if we have things configured the way we hope, right? So, maybe this doesn't have to be listed as a breaking change. I'll let you decide. However, calling it a breaking change, but not using a breaking change major version bump is unnecessarily inconsistent.

docs/decisions/0009-course-catalog-info-changed-design.rst
```diff
@@ -11,19 +11,23 @@ Context
 
 The COURSE_CATALOG_INFO_CHANGED signal is used by Studio to convey that information relevant to the course catalog has changed. It was created with an eye towards eventually replacing the refresh-course-metadata batch job, which syncs all data between edx-platform and course-discovery.
 
-Refresh-course-metadata functions differently depending on whether or not the system has Publisher enabled. Specifically, after requesting information from the ``/courses`` endpoint in Studio, Discovery will ignore certain fields if Publisher is enabled. These fields are mostly part of the ``media`` attribute of the API response. They are sent in a variety of ways (absolute urls, paths, sometimes divided by size, etc.)
+Refresh-course-metadata functions differently depending on whether or not the system has Publisher enabled. Specifically, after requesting information from the ``/courses`` endpoint in Studio, Discovery will ignore certain fields if Publisher is enabled. Most, thought not all, of these fields are sent as part of the ``media`` attribute of the API response. These media fields are sent in a variety of ways (absolute urls, paths, sometimes divided by size, etc.)
```
robrap, line 14: typo: thought => though

docs/decisions/0009-course-catalog-info-changed-design.rst
```diff
@@ -11,19 +11,23 @@ Context
...
 
 Decision
 --------
 
-The COURSE_CATALOG_INFO_CHANGED will only contain the information necessary to work in a Publisher-enabled environment. In particular, this means it will not contain the ``media`` field usually present in the Studio ``/courses`` API endpoint.
+The COURSE_CATALOG_INFO_CHANGED will only contain the information necessary to work in a Publisher-enabled environment. In particular, this means it will not contain the following fields usually present in the Studio ``/courses`` API endpoint:
```
robrap, line 19: Not sure if you'll want to use language like "some example fields", or "as of the writing of this ADR". I don't think this is a list we'll want to maintain if it ever changes.

docs/decisions/0009-course-catalog-info-changed-design.rst
```diff
@@ -11,19 +11,23 @@ Context
...
 
 In Discovery, if Publisher is not enabled, the consumer will log a warning and not try to update anything.
 
 Rationale
 ---------
 
-The way we update media information in refresh-course-metadata is quite haphazard, with no real standard way of sending over the information or of storing it on the Discovery end. Replicating these structures in the COURSE_CATALOG_INFO_CHANGED signal would make the data definition confusing.
+The way we update media information in refresh-course-metadata is quite haphazard, with no real standard way of sending over the information or of storing it on the Discovery end. Replicating these structures in the COURSE_CATALOG_INFO_CHANGED signal would make the data definition confusing. In addition, it is very difficult to test these other fields in the edX.org staging and production environments because, since publisher is enabled, these fields are not present in studio.
 
 Until these fields are needed by a consumer in a non-Publisher environment (if ever), it makes sense to defer trying to find a good solution to this problem.
```
robrap, line 32: I would move the new sentence from above ("In addition, it is very difficult...") to this section. I think this sentence adds better context. Also, I'd try to avoid edx.org references if we can. Something like:
```suggestion
Until these fields are needed by a consumer in a non-Publisher environment (if ever), it makes sense to defer trying to find a good solution to this problem. Additionally, it is very difficult to test this without a real non-Publisher environment that is running refresh-course-metadata.
```

robrap (commented):
Thanks.

CHANGELOG.rst
rgraber: This did end up being a breaking change because of how python named arguments work, interestingly. If you call CourseCatalogData(..., effort=5) the code will error. I'll fix the version. This certainly raises some questions for our future work, but for now I'll just call it a breaking change and deal.

rgraber: ```TypeError: __init__() got an unexpected keyword argument 'effort'``` was the actual error. I'm actually not sure why we didn't hit this with other fields unless we removed them before we actually tried using them in Studio. Doing a quick code hunt to see.

rgraber:
Ended up rebasing/squashing to make sure my commit messages were all compliant, and because I forgot that would make your life slightly harder. I made all suggested changes to the ADR, changed the version to 3.0.0, and updated the Changelog. Those should be the only changes since your last review.

docs/decisions/0009-course-catalog-info-changed-design.rst
rgraber: Fixed (locally, not with commit suggestion)

CHANGELOG.rst
rgraber: Fixed: updated to 3.0.0 and noted that the field should not actually be in use

docs/decisions/0009-course-catalog-info-changed-design.rst
rgraber: fixed

docs/decisions/0009-course-catalog-info-changed-design.rst
rgraber: Fixed: used "for example"

robrap (approved):

PR #132, opened 2022-10-18
fix: Increase unit test coverage (string_schema, format_responses)

timmc-edx:
This includes a small change to logging to make it easier to test.

**Merge checklist:**
- [ ] Version bumped
- [ ] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/openedx_events/tests/test_tooling.py
+++ b/openedx_events/tests/test_tooling.py
@@ -4,6 +4,7 @@ This file contains all test for the tooling.py file.
 Classes:
     EventsToolingTest: Test events tooling.
 """
+from contextlib import contextmanager
 from unittest.mock import Mock, patch
 
 import attr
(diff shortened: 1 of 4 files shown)


timmc-edx:
I set out to improve coverage, but it turns out most of the missing coverage is in https://github.com/openedx/openedx-events/blob/main/openedx_events/tests/utils.py -- which is all test utilities that maybe we don't care about coverage for. :-) I'd be happy to add a no-cover pragma if you think it makes sense.

openedx_events/tests/test_tooling.py
```diff
@@ -37,6 +53,14 @@ def setUp(self):
             data=self.data_attr,
         )
 
+        self.receiver_error = Exception("fake error")
+
+        def error_receiver(*args, **kwargs):
+            raise self.receiver_error
+
+        self.ok_receiver = Mock(return_value="success")
+        self.error_receiver = error_receiver
```
timmc-edx, line 62: Author note: These test receivers and `receivers_attached` could be used in other tests in this file, which is why they're up here. If you don't think we'll get around to using them in other tests, I can collapse them back down into the one place they're used.

timmc-edx:
I also noticed that `InstantiationError` is no longer referenced, having previously been removed in #23. I don't see references elsewhere in edx or openedx. Can probably remove safely. (It's responsible for a few lines of coverage loss.)

openedx_events/tests/test_tooling.py
mariajgrimaldi: This looks really cool! And I see it can be used elsewhere if needed

mariajgrimaldi (approved):

PR #138, opened 2022-10-31
fix: use None instead of null as default value for optional fields

rgraber:
**Description:** 
Use None instead of null as the default value for optional fields to fix the fastavro error "Default value <null> must match first schema in union with type: null"

See https://github.com/openedx/openedx-events/pull/137

**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ----------
 
+[3.0.1] - 2022-10-31
+--------------------
+Fixed
+~~~~~
+* Fix default value for optional fields from "null" to None
+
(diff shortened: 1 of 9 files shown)


dianakhuang (approved):

felipemontoya (approved):
This looks good. Thanks for the speed to produce the fix

robrap (approved):

PR #139, opened 2022-10-31
fix: contrain fastavro

robrap:
**Description:** 

Add constraint for fastavro, forcing manual
upgrades to ensure it passes tests before
being upgraded in dependent services.

**ISSUE:**

Here was a PR related to a fastavro change that broken unit tests: https://github.com/openedx/openedx-events/pull/137. However, because we weren't constraining, services had upgraded the breaking change. This would be most dangerous if it started breaking event production.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped (**this is still needed!**)
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

My concern is that this means this constraint would need to be manually changed any time we wanted to upgrade. Who would watch for this and ensure we are getting security updates, or not getting too far behind, etc.?

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -8,3 +8,8 @@
 # pin when possible.  Writing an issue against the offending project and
 # linking to it here is good.
 -c common_constraints.txt
+
+# Keep fastavro constrained to ensure that unit tests pass before dependent services
+#  try to update fastavro. Upgrade at any time by adjusting to the latest version,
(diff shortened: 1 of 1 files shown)


robrap:
Closing this example PR until https://github.com/openedx/openedx-events/issues/140 is investigated further, because we may or may not want this.

PR #158, opened 2022-12-14
[FC-0004][RG]/edxoldmng 218/Signals to save/delete certificate configuration for a course.

NikolayBorovenskiy:
**Description:** This pr adds signals to save/delete certificate configuration for a course.

**ISSUE:** [Link to GitHub issue](https://github.com/openedx/credentials/issues/1765)

**Reviewers:**
- [ ] @ormsbee 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/openedx_events/content_authoring/data.py
+++ b/openedx_events/content_authoring/data.py
@@ -8,6 +8,7 @@ The attributes for the events come from the CourseDetailView in the LMS, with so
 (see deprecation proposal at https://github.com/openedx/public-engineering/issues/160)
 """
 from datetime import datetime
+from typing import BinaryIO, List
 
 import attr
 from opaque_keys.edx.keys import CourseKey, UsageKey
@@ -82,3 +83,56 @@ class DuplicatedXBlockData(XBlockData):
     """
 
     source_usage_key = attr.ib(type=UsageKey)
+
+
+@attr.s(frozen=True)
+class CertificateSignatoryData:
+    """
+    Attributes defined for Open edX CertificateSignatory data object.
+
+    Arguments:
+        image (BinaryIO): certificate signature image.
+        name (str): name of signatory.
+        organization (str): organization that signatory belongs to.
+        title (int): signatory title.
+    """
+
+    # Note: Please take care that the image field is BinaryIO, which means
+    # that a file can be passed as an array of bytes. Watch the size of this file.
+    # It can potentially be large, making it difficult to pass this data structure through the Event Bus
+    # (CloudEvent messages, which should be 64K or less) and store it on disk space.
+    # We suggest referring to MAX_ASSET_UPLOAD_FILE_SIZE_IN_MB, i.e. restriction in the Studio for such cases.
+    image = attr.ib(type=BinaryIO)
+    # end Note
+    name = attr.ib(type=str)
+    organization = attr.ib(type=str)
+    title = attr.ib(type=str)
+
+
+@attr.s(frozen=True)
+class CertificateConfigData:
+    """
+    Attributes defined for Open edX CertificateConfig data object.
+
+    Arguments:
+        certificate_type (str): certificate type. Possible types are certificate relevant course modes:
+         - credit,
+         - verified,
+         - professional,
+         - no-id-professional,
+         - executive-education,
+         - paid-executive-education,
+         - paid-bootcamp,
+         - masters.
+        course_key (CourseKey): identifier of the Course object.
+        title (str): certificate title.
+        signatories (List[CertificateSignatoryData]): contains a collection of signatures
+        that belong to the certificate configuration.
+        is_active (bool): indicates whether the certifivate configuration is active.
+    """
+
+    certificate_type = attr.ib(type=str)
+    course_key = attr.ib(type=CourseKey)
+    title = attr.ib(type=str)
+    signatories = attr.ib(type=List[CertificateSignatoryData], factory=list)
+    is_active = attr.ib(type=bool, default=False)
--- a/openedx_events/content_authoring/signals.py
+++ b/openedx_events/content_authoring/signals.py
@@ -8,7 +8,12 @@ They also must comply with the payload definition specified in
 docs/decisions/0003-events-payload.rst
 """
 
-from openedx_events.content_authoring.data import CourseCatalogData, DuplicatedXBlockData, XBlockData
+from openedx_events.content_authoring.data import (
+    CertificateConfigData,
+    CourseCatalogData,
+    DuplicatedXBlockData,
+    XBlockData,
+)
 from openedx_events.tooling import OpenEdxPublicSignal
 
 # .. event_type: org.openedx.content_authoring.course.catalog_info.changed.v1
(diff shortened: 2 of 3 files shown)


mariajgrimaldi:
@mphilbrick211: hello 😄, what should we do with the CLA issue here?

mariajgrimaldi:
@mphilbrick211: hello 😄, what should we do with the CLA issue here?

openedx_events/content_authoring/data.py
```diff
@@ -54,3 +54,42 @@ class CourseCatalogData:
...
+    Attributes defined for Open edX CertificateSignatory data object.
+
+    Arguments:
+        image (BinaryIO): certificate signature image.
+        name (str): name of signatory.
+        organization (str): organization that signatory belongs to.
+        title (int): signatory title.
+    """
+
+    image = attr.ib(type=BinaryIO)
```
ormsbee, line 71: Can this be a URL instead (`image_url`)? Signals will get translated to CloudEvent messages, which should be 64K or less. There's no guarantee that a professor's signature will fit in that limit.

openedx_events/content_authoring/data.py
```diff
@@ -54,3 +54,42 @@ class CourseCatalogData:
...
+        certificate_type (str): certificate type.
+        course_id (str): identifier of the course to which the certificate config belongs.
+        title (str): certificate title.
+        signatories (List[CertificateSignatoryData]): contains a collection of signatures
+        that belong to the certificate configuration.
+        is_active (bool): indicates whether the certifivate configuration is active.
+    """
+
+    certificate_type = attr.ib(type=str)
+    course_id = attr.ib(type=str)
```
ormsbee, line 92: In keeping with other conventions in this repo, please call this `course_key` and make it of type `CourseKey`.

openedx_events/content_authoring/data.py
```diff
@@ -54,3 +54,42 @@ class CourseCatalogData:
...
+    Arguments:
+        certificate_type (str): certificate type.
+        course_id (str): identifier of the course to which the certificate config belongs.
+        title (str): certificate title.
+        signatories (List[CertificateSignatoryData]): contains a collection of signatures
+        that belong to the certificate configuration.
+        is_active (bool): indicates whether the certifivate configuration is active.
+    """
+
+    certificate_type = attr.ib(type=str)
```
ormsbee, line 91: In the docstring, please either enumerate the types possible, or describe where the type field data comes from and give examples.

openedx_events/content_authoring/signals.py
```diff
@@ -21,3 +21,26 @@
...
+)
+
+# .. event_type: org.openedx.content_authoring.course.certificate_config.deleted.v1
+# .. event_name: COURSE_CERTIFICATE_CONFIG_DELETED
+# .. event_description: Fired when a course certificate configuration deletes in Studio.
+# .. event_data: CertificateConfigData
+COURSE_CERTIFICATE_CONFIG_DELETED = OpenEdxPublicSignal(
+    event_type="org.openedx.content_authoring.course.certificate_config.deleted.v1",
+    data={
+        "certificate_config": CertificateConfigData,
```
ormsbee, line 44: Why do we need to pass all this information if it's been deleted? Can we just pass the `course_key` instead?

ormsbee (changes requested):

mphilbrick211:
> @mphilbrick211: hello 😄, what should we do with the CLA issue here?

Thanks for the ping, @mariajgrimaldi! We're looking into this now, and hope to be able to push it through as soon as we can.

openedx_events/content_authoring/data.py
NikolayBorovenskiy: This is a very good question and I thought about it. 
Why did I do this?
First, this is not the final version yet. What we are doing is an intermediate stage of moving towards where the configuration of certificates, the certificates themselves will be located in Credentials service.
Now the signature files are saved in MongoDB as an asset and passed to Credentials via HTTP. Also, it is passed as an ordinary multipart request where files are sent as bytes. This is the usual approach for sending files. Yes, now we use Event and look to the future, we are thinking about the future architecture, where there is a message Event Bus and events. We are doing our best to make it easy to switch to this, but for now, we are temporarily passing the certificate configuration through the HTTP API. So far, as we agreed before.

For this to be `image_url`, then, when saving the configuration from Studio UI, you first need to save the file somewhere on a file system or in a cloud and pass it to the Credentials service, and this service itself will also have access to this shared storage. When we do not have Mongo DB for cert configs and signatures, and we keep this in mind, then these certificate configuration files will not be saved there and they need to be passed immediately to Credentials. Then in order for it to be `image_url`, it must first be saved somewhere in intermediate public storage. But the problem is that I do not yet know what it will be and what architectural solution will be applied. This is not known. Since I wrote at the very beginning that we are at an intermediate stage of transition.  

What you're talking about, where size matters is easy to solve with server-side validation, and avoid where you need to transfer a large file. Or do post-processing of the file before passing this information through the bus. But all this will be decided and a decision will be made on how to do it a little later, at the next stage, so to speak, when more becomes known and when the transfer of the certificate configuration will be completely through EvenBus, and not imitation as we are doing now. That's why for now I propose to leave it as it is, but keep in mind those solutions and problems that I described. 

What do you think?

openedx_events/content_authoring/signals.py
NikolayBorovenskiy: I think we should. Why did I do this? It's all about the low coupling of the two services Studio and Credentials. We have to support it and not depend on each other. How can we efficiently delete resources in general? How do we do it with a regular http request in monolithic systems, i.e. explicitly pass an entity identifier, which is usually the primary key in a database table, or pass some unique field at the database level. Then the removal will be accurate and correct.

It’s not about our case now. Studio does not store the ID and does not know about the model from the Credentials database. It's good. course_key may or may not be a unique field. Studio doesn't know this for sure and shouldn't know. Only Credentials knows this. What can we do then when we keep services loosely coupled? Pass on as much meaningful information as possible and ask for Credentials to be removed. And Credentials itself will decide which record in the table needs to be deleted, since it is an information expert here. Which is what we do.

openedx_events/content_authoring/data.py
NikolayBorovenskiy: Good. I'll do it. Thank you.

openedx_events/content_authoring/data.py
NikolayBorovenskiy: Got it.

mphilbrick211:
Hi @NikolayBorovenskiy! When you re-run the checks, your CLA should be all set and the check should turn green. Please let me know if you have any questions.

NikolayBorovenskiy:
Hi, @mphilbrick211  looks like that CLA is good now. Thanks for pushing me.

openedx_events/content_authoring/data.py
ormsbee: My comments had two assumptions:

1. Backwards compatibility concerns will require us to continue to support whatever fields are added over the long term. In order to not break plugins in the future, openedx-events opts for a [pretty strict conventions](https://github.com/openedx/openedx-events/blob/main/docs/decisions/0006-event-schema-serialization-and-evolution.rst) around schema evolution.
2. There is already a publicly accessible image URL representing the signature, since we need to display that signature to users–so we wouldn't have to save the file anywhere.

openedx_events/content_authoring/signals.py
ormsbee: > course_key may or may not be a unique field.

But there's only ever one credentials configuration for a given course_key, right? When would we ever have more than one?

NikolayBorovenskiy: Not certainly in that way. Look at this model please https://github.com/openedx/credentials/blob/master/credentials/apps/credentials/models.py#L194-L230
Here course_id is just a CharField and not unique. And here is our constraint `unique_together = (("course_id", "certificate_type", "site")),)` which explains the behavior.

And in any case, it's not so important. After all, as I wrote above, Studio is not an information expert for Credentials and does not know the exact URI of this resource. It is important for us to maintain low coupling and sending more data will not be superfluous here.

openedx_events/content_authoring/data.py
NikolayBorovenskiy: Dave, I suggest we take a step back if you don't mind. I want us to be on the same page and I'm afraid to take sharp, thoughtless steps.
Since we are talking about implementation, I suggest going to this PR https://github.com/openedx/edx-platform/pull/31446. This is where the implementation is. You will be able to understand what was done and how. We will approve the final decision, and then we will continue our work here.

What do you say?

openedx_events/content_authoring/signals.py
ormsbee: I see. I misunderstood in that I thought CertificateConfigData was the representation of all certificates for a given course (i.e. a cleaner version of the entire `certificates` attribute off of the CourseBlock), rather than being the configuration for one particular certificate type for a course.

openedx_events/content_authoring/data.py
ormsbee: All right, let's just see how this goes. You can keep the BinaryIO image, but please log how large they get in the code that constructs these things so that we can examine that data after edx.org runs it.

ormsbee (approved):

openedx_events/content_authoring/data.py
NikolayBorovenskiy: I'll do.

NikolayBorovenskiy:
@ormsbee Dave, I've pushed all we need. Please verify this. Tnx in advance a lot.

ormsbee:
@NikolayBorovenskiy: Please squash and fix the quality error (sort order of imports) and I'll merge.

NikolayBorovenskiy:
@ormsbee Dave, thanks a lot. I've done my part. Please do a final review and if the checkers are fine this time let's merge and make a new tag.

timmc-edx:
Would you mind making a quick PR to add this to the changelog as well?

timmc-edx:
Actually, if you could just review https://github.com/openedx/openedx-events/pull/180 that would be great -- let me know if that summary is correct or not.

Issue #162, opened 2022-12-23
Discovery: Signal metadata, event bus consumers, and timestamps


robrap:
We recently ran into an issue where we want our event bus signal consumer logic to know the broker's timestamp of the message. However, that is not available in the signal handler, and brings up a number of issues:

1. If an event bus signal consumer looked at the "metadata" dict in the kwargs, what data would it find? Some valid and some invalid data? All invalid? Other?
  * Note: The signal currently has a `time` metadata with `now()` as the implementation, which would give the consumer time.  
3. Is it safe to assume that the metadata on the consumer _should_ match what would be seen on the producer? How should this data be sent, and where should the metadata be updated? Note that most of this data is sent in the message headers as a start.
4. If we want to send event bus specific data to the signal handler, like the event bus broker's timestamp for the message, how would we do this?
5. Is there a way to not be blocked on https://github.com/openedx/openedx-events/issues/159 to find a single definition of "time"? Can we avoid that field, and use more specifically named and defined fields?

**A/C:**
- [x] Ensure the consumer has access to the time when an event was produced
- [x] Ensure the consumer can send the original event metadata when it emits a signal (or at the very least, the signal handler has access to that data)
- [x] Update docs accordingly

**Additional tasks:**
- [x] openedx-events requires update to accept time.
- [x] openedx-events requires update for `send_event_with_metadata`.
- [x] edx-platform will require an update to send the db time from mongo. [Note: Robert started this locally.]
- [x] event-bus-kafka needs several changes:
   - [x] Add ce_time, sourcelib, and minorversion to header.
   - [x] Add new header items to logs, unless we get that for free. [Note: We do indeed get this for free]
   - [x] Switch to call `send_event_with_metadata` for consumers.

robrap:
Note: discovery needs a temporary hack for delaying event processing, and we need to know the event timestamp to know whether or not to delay.

robrap:
Here are docs for Kafka timestamp, which may be from producer or broker: https://docs.confluent.io/5.0.0/clients/confluent-kafka-dotnet/api/Confluent.Kafka.Timestamp.html#Confluent_Kafka_Timestamp__ctor_System_Int64_Confluent_Kafka_TimestampType_

I should also link to the Phoenix ticket for the UUID concurrency issue, which is the issue that needed this data.

ormsbee:
[...]

PR #168, opened 2023-01-17
feat: updated event signal metadata time

robrap:
**Description:**

* Updated time metadata to include UTC timezone. Original implementation used utcnow(), which could give different results if the time were ever
interpreted to be local time. See https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow
* Updated send_event to optionally take a time, which would be used for the metadata. Ideally, this will be explicitly set.

**ISSUE:**

* Implements first part of https://github.com/openedx/openedx-events/issues/162
* edx-platform will require an update to send the db time for the catalog event.
* event-bus-kafka will require passing on the time during event production.  
* Another change is needed in this library for `send_event_in_bus_consumer`, which will require an additional breaking change to EventsMetadata. That will likely be made in a separate branch and merged into this PR.

**Testing instructions:**

* Unit tests are probably enough for this change.
* The edx-platform and event-bus-kafka changes that will use this will be used to provide more end-to-end testing.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] ~~Version bumped~~
- [x] Changelog record added
- [x] Commits are squashed
- [x] Remove breaking change indicator from commit message

**Post merge:**
- [x] ~~Create a tag~~
- [x] ~~Check new version is pushed to PyPI after tag-triggered build is
      finished.~~

**Author concerns:** 

I don't love how EventsMetadata is initialized with time that can be None, but this is how I think it needs to be done for the frozen attributes. Note, this pattern is going to be followed in the follow-up PR where all arguments need to be sent for event bus consumers.

### DIFF
--- a/openedx_events/data.py
+++ b/openedx_events/data.py
@@ -42,7 +42,9 @@ class EventsMetadata:
         time (datetime): (optional) timestamp when the event was sent with
             UTC timezone. Defaults to current time in UTC. See OEP-41 for
             details.
-        sourcelib (str): Open edX Events library version.
+        sourcelib (tuple of ints): Open edX Events library version. A tuple was
+            selected so that version comparisons don't have to worry about
+            lexical ordering of strings (e.g. '0.9.0' vs. '0.10.0').
     """
 
     id = attr.ib(type=UUID, init=False)
--- a/openedx_events/data.py
+++ b/openedx_events/data.py
@@ -5,13 +5,24 @@ These attributes follow the form of attr objects specified in OEP-49 data
 pattern.
 """
 import socket
-from datetime import datetime
+from datetime import datetime, timezone
 from uuid import UUID, uuid1
 
 import attr
 from django.conf import settings
 
 import openedx_events
+
+
+def _ensure_utc_time(_, attribute, value):
+    """
+    Ensure the value is a UTC datetime.
+
+    Note: Meant to be used along-side an instance_of attr validator.
+    """
+    if value.tzinfo and value.tzinfo == timezone.utc:
+        return
+    raise ValueError(f"'{attribute.name}' must have timezone.utc")
 
 
 @attr.s(frozen=True)
@@ -28,7 +39,9 @@ class EventsMetadata:
         minorversion (int): version of the event type.
         source (str): logical source of an event.
         sourcehost (str): physical source of the event.
-        time (datetime): timestamp when the event was sent.
+        time (datetime): (optional) timestamp when the event was sent with
+            UTC timezone. Defaults to current time in UTC. See OEP-41 for
+            details.
         sourcelib (str): Open edX Events library version.
     """
 
@@ -37,7 +50,13 @@ class EventsMetadata:
     minorversion = attr.ib(type=int, converter=attr.converters.default_if_none(0))
     source = attr.ib(type=str, init=False)
     sourcehost = attr.ib(type=str, init=False)
-    time = attr.ib(type=datetime, init=False)
+    current_utc_time = datetime.now(timezone.utc)
+    time = attr.ib(
+        type=datetime,
+        default=None,
+        converter=attr.converters.default_if_none(attr.Factory(lambda: datetime.now(timezone.utc))),
+        validator=attr.validators.optional([attr.validators.instance_of(datetime), _ensure_utc_time]),
+    )
     sourcelib = attr.ib(type=tuple, init=False)
 
     def __attrs_post_init__(self):
@@ -56,7 +75,6 @@ class EventsMetadata:
             ),
         )
         object.__setattr__(self, "sourcehost", socket.gethostname())
-        object.__setattr__(self, "time", datetime.utcnow())
         object.__setattr__(
             self, "sourcelib", tuple(map(int, openedx_events.__version__.split(".")))
         )
--- a/openedx_events/tests/test_tooling.py
+++ b/openedx_events/tests/test_tooling.py
@@ -4,8 +4,10 @@ This file contains all test for the tooling.py file.
 Classes:
     EventsToolingTest: Test events tooling.
 """
+import datetime
 from contextlib import contextmanager
 from unittest.mock import Mock, patch
+from uuid import UUID
 
 import attr
 import ddt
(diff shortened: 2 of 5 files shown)


openedx_events/data.py
```diff
@@ -56,7 +72,8 @@ def __attrs_post_init__(self):
             ),
         )
         object.__setattr__(self, "sourcehost", socket.gethostname())
-        object.__setattr__(self, "time", datetime.utcnow())
         object.__setattr__(
             self, "sourcelib", tuple(map(int, openedx_events.__version__.split(".")))
         )
+        if getattr(self, "time", None) is None:
```
rgraber, line 78: Can we use the same default_if_none converter used in minorversion?

rgraber (commented):
A few questions

openedx_events/data.py
```diff
@@ -37,7 +48,12 @@ class EventsMetadata:
     minorversion = attr.ib(type=int, converter=attr.converters.default_if_none(0))
     source = attr.ib(type=str, init=False)
     sourcehost = attr.ib(type=str, init=False)
-    time = attr.ib(type=datetime, init=False)
+    current_utc_time = datetime.now(timezone.utc)
+    time = attr.ib(
+        type=datetime,
+        # time is not actually optional, but gets filled in post init if None
```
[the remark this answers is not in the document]
robrap: I dropped the confusing comment. I meant something more like: "This argument is required, even though the value is optional". But, better to not say anything I think.

openedx_events/data.py
robrap: We cannot. I added a comment to explain.

openedx_events/tests/test_tooling.py
```diff
@@ -101,10 +103,58 @@ def test_get_signal_metadata(self, socket_mock, events_package_mock):
             "sourcehost": "edx.devstack.lms",
             "sourcelib": [0, 1, 0],
         }
+        near_expected_time = datetime.datetime.now(datetime.timezone.utc)
 
         metadata = self.public_signal.generate_signal_metadata()
 
         self.assertDictContainsSubset(expected_metadata, attr.asdict(metadata))
+        self.assertIsInstance(metadata.id, UUID)
+        assert abs((metadata.time - near_expected_time).seconds) <= 1
```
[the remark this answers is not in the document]
robrap: I agree, and found a better way.

openedx_events/tests/test_tooling.py
```diff
@@ -155,6 +205,30 @@ def test_send_robust_event_successfully(self, format_responses_mock, log_mock, f
             "Responses of the Open edX Event <org.openedx.learning.session.login.completed.v1>: \nfake-output"
         )
 
+    @patch("openedx_events.tooling.OpenEdxPublicSignal.generate_signal_metadata")
+    @patch("openedx_events.tooling.Signal.send")
+    def test_send_event_with_time_successfully(self, send_mock, fake_metadata):
+        """
+        This method tests the process of sending an event that's allow to fail.
```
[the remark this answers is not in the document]
robrap: 1. I fixed the original typo.
2. I took a little more care to update this test based on its purpose, rather than leaving copy-pasted cruft.

openedx_events/data.py
```diff
@@ -37,7 +48,12 @@ class EventsMetadata:
     minorversion = attr.ib(type=int, converter=attr.converters.default_if_none(0))
     source = attr.ib(type=str, init=False)
     sourcehost = attr.ib(type=str, init=False)
-    time = attr.ib(type=datetime, init=False)
+    current_utc_time = datetime.now(timezone.utc)
+    # Note: cannot use default_if_none for time, because it caches the time the class is loaded.
+    time = attr.ib(
```
rgraber, line 53: Can we do something like 
```
time = attr.ib(
    type=datetime,
    default = attr.Factory(lambda: datetime.now(timezone.utc), takes_self=False),
    validator=attr.validators.optional([attr.validators.instance_of(datetime), _ensure_utc_time]),
)
```
?
Asking because I think future us might be confused and/or annoyed at constantly having to put time=None on things. FWIW I tried something like this locally and because it uses a callable it didn't cache the time.

rgraber (commented):
One more thought about defaults

openedx_events/data.py
rgraber: (and honestly if I had my way we'd do the same thing for minorversion)

openedx_events/tests/test_tooling.py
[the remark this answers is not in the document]
rgraber: Possible missed commit? I still see the old test

openedx_events/data.py
robrap: I found the right combo. I didn't know about `attr.Factory`, and had just tried passing `default_if_none` a lamba orginally, and that didn't work. Passing it the factory works.

openedx_events/data.py
```diff
@@ -28,16 +39,26 @@ class EventsMetadata:
...
         source (str): logical source of an event.
         sourcehost (str): physical source of the event.
-        time (datetime): timestamp when the event was sent.
-        sourcelib (str): Open edX Events library version.
+        time (datetime): (optional) timestamp when the event was sent with
+            UTC timezone. Defaults to current time in UTC. See OEP-41 for
+            details.
+        sourcelib (tuple of ints): Open edX Events library version. A tuple was
+            selected so that version comparisons don't have to worry about
+            lexical ordering of strings (e.g. '0.9.0' vs. '0.10.0').
```
robrap, line 47: @rgraber: This is in its own commit, but sorry about adding it to this PR. The original type was wrong, and I wasn't sure why we chose tuples. The comment comes from https://github.com/openedx/openedx-events/pull/7#discussion_r667294019.

openedx_events/data.py
robrap: We can, and you helped me learn how.

openedx_events/data.py
rgraber: fancy nesting! thanks

rgraber (approved):
LGTM, thanks for the updates

PR #172, opened 2023-01-20
fix: make minorversion actually optional

robrap:
**Description:**

You can now create an EventsMetadata instance
without supplying minorversion, and get the
default of 0.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] ~~Version bumped~~
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [x] ~~Create a tag~~
- [x] ~~Check new version is pushed to PyPI after tag-triggered build is
      finished.~~
- [x] Delete working branch (if not needed anymore)

### DIFF
--- a/openedx_events/data.py
+++ b/openedx_events/data.py
@@ -36,7 +36,7 @@ class EventsMetadata:
     Arguments:
         id (UUID): event identifier.
         event_type (str): name of the event.
-        minorversion (int): version of the event type.
+        minorversion (int): (optional) version of the event type. Defaults to 0.
         source (str): logical source of an event.
         sourcehost (str): physical source of the event.
         time (datetime): (optional) timestamp when the event was sent with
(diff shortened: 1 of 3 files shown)


openedx_events/data.py
```diff
@@ -34,7 +34,7 @@ class EventsMetadata:
 
     id = attr.ib(type=UUID, init=False)
     event_type = attr.ib(type=str)
-    minorversion = attr.ib(type=int, converter=attr.converters.default_if_none(0))
+    minorversion = attr.ib(type=int, default=None, converter=attr.converters.default_if_none(0))
```
robrap, line 37: @mariajgrimaldi: Since we already had a `default_if_none` converter, and we could have used `EventsMetadata(..., minorversion=None)`, I presume it is acceptable to not require it to be supplied at all? Please confirm that this wasn't because we want to force it as an argument, in which case we should probably go the other way and remove `default_if_none` or something. Let me know what you think of this PR. Thanks!

mariajgrimaldi: You're right! We don't need it here since we have the converter :)

rgraber (approved):
LGTM

PR #175, opened 2023-01-24
feat: add send_event_with_custom_metadata

robrap:
**Description:**

Added send_event_with_custom_metadata. This new version of send_event enables event bus consumers to send the event signal with the same metadata fields that were used when the event was produced.

**ISSUE:**

https://github.com/openedx/openedx-events/issues/162

**Testing instructions:**

Only tested via unit tests at this point. I could possibly make other PRs using these changes and pulling in openedx-events from this branch. But, it may be simpler to just release and fix forward if there are any issues, since no one else will be using the new method.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed (update comment with new name send_event_with_custom_metadata)

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/openedx_events/data.py
+++ b/openedx_events/data.py
@@ -47,36 +47,39 @@ class EventsMetadata:
             lexical ordering of strings (e.g. '0.9.0' vs. '0.10.0').
     """
 
-    id = attr.ib(type=UUID, init=False)
-    event_type = attr.ib(type=str)
-    minorversion = attr.ib(type=int, default=None, converter=attr.converters.default_if_none(0))
-    source = attr.ib(type=str, init=False)
-    sourcehost = attr.ib(type=str, init=False)
-    current_utc_time = datetime.now(timezone.utc)
-    time = attr.ib(
-        type=datetime,
-        default=None,
-        converter=attr.converters.default_if_none(attr.Factory(lambda: datetime.now(timezone.utc))),
-        validator=attr.validators.optional([attr.validators.instance_of(datetime), _ensure_utc_time]),
+    event_type = attr.ib(type=str, validator=attr.validators.instance_of(str))
+    id = attr.ib(
+        type=UUID, default=None,
+        converter=attr.converters.default_if_none(attr.Factory(lambda: uuid1())),  # pylint: disable=unnecessary-lambda
+        validator=attr.validators.instance_of(UUID),
     )
-    sourcelib = attr.ib(type=tuple, init=False)
-
-    def __attrs_post_init__(self):
-        """
-        Post-init hook that generates metadata for the Open edX Event.
-        """
-        # Have to use this to get around the fact that the class is frozen
-        # (which we almost always want, but not while we're initializing it).
-        # Taken from edX Learning Sequences data file.
-        object.__setattr__(self, "id", uuid1())
-        object.__setattr__(
-            self,
-            "source",
-            "openedx/{service}/web".format(
-                service=getattr(settings, "SERVICE_VARIANT", "")
-            ),
-        )
-        object.__setattr__(self, "sourcehost", socket.gethostname())
-        object.__setattr__(
-            self, "sourcelib", tuple(map(int, openedx_events.__version__.split(".")))
-        )
+    minorversion = attr.ib(
+        type=int, default=None,
+        converter=attr.converters.default_if_none(0), validator=attr.validators.instance_of(int)
+    )
+    source = attr.ib(
+        type=str, default=None,
+        converter=attr.converters.default_if_none(
+            attr.Factory(lambda: "openedx/{service}/web".format(service=getattr(settings, "SERVICE_VARIANT", "")))
+        ),
+        validator=attr.validators.instance_of(str),
+    )
+    sourcehost = attr.ib(
+        type=str, default=None,
+        converter=attr.converters.default_if_none(
+            attr.Factory(lambda: socket.gethostname())  # pylint: disable=unnecessary-lambda
+        ),
+        validator=attr.validators.instance_of(str),
+    )
+    time = attr.ib(
+        type=datetime, default=None,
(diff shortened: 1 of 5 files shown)


openedx_events/data.py
```diff
@@ -47,36 +47,36 @@ class EventsMetadata:
             lexical ordering of strings (e.g. '0.9.0' vs. '0.10.0').
     """
 
-    id = attr.ib(type=UUID, init=False)
+    id = attr.ib(
+        type=UUID, default=None,
+        converter=attr.converters.default_if_none(attr.Factory(lambda: uuid1())),  # pylint: disable=unnecessary-lambda
```
robrap, line 52: **Note to Reviewer**: 
1. The pylint disable is because it was simpler to mock and test with a lambda.
2. The following `default_if_none` converters were all based on code that was in `__attrs_post_init__`, but was removed in this PR.

Please mark resolved once read by main reviewer.

openedx_events/tooling.py
```diff
@@ -185,6 +151,81 @@ def validate_sender():
...
+        Exceptions raised:
+            SenderValidationError: raised when there's a mismatch between
+            arguments passed to this method and arguments used to initialize
+            the event.
+        """
+        metadata = self.generate_signal_metadata(time=time)
+        return self._send_event_with_metadata(metadata=metadata, send_robust=send_robust, **kwargs)
+
+    def send_event_in_bus_consumer(
+        self, id, minorversion, source, sourcehost, time, sourcelib,  # pylint: disable=redefined-builtin
```
robrap, line 198: **Note to Reviewer**: The pylint disable is because I wanted to use `id`, to match the name used in EventsMetadata.
Please mark resolved once read by main reviewer.

openedx_events/data.py
```diff
@@ -47,36 +47,36 @@ class EventsMetadata:
...
+        validator=attr.validators.instance_of(UUID),
```
robrap, line 53: **Note to Reviewer**: I also added type validators to ensure the data being sent is appropriate.
Please mark resolved once read by main reviewer.

openedx_events/tooling.py
```diff
@@ -185,6 +151,81 @@ def validate_sender():
 
         return responses
 
+    def send_event(self, send_robust=True, time=None, **kwargs):
```
robrap, line 154: Unfortunately, time is optional and after send_robust, which is the opposite order of the arguments in the new method, to keep this as backward compatible as possible.

openedx_events/data.py
rgraber: [question] does type=UUID not do this for us? same for all the other updated fields below

openedx_events/tooling.py
timmc-edx, line 198: There are a ton of params, so can we make these kwarg-only? I think it would be like so:

```suggestion
        self, *, id, minorversion, source, sourcehost, time, sourcelib,  # pylint: disable=redefined-builtin
```

openedx_events/tooling.py
```diff
@@ -185,6 +151,81 @@ def validate_sender():
...
+
+        Exceptions raised:
+            SenderValidationError: raised when there's a mismatch between
+            arguments passed to this method and arguments used to initialize
+            the event.
+        """
+        metadata = self.generate_signal_metadata(time=time)
+        return self._send_event_with_metadata(metadata=metadata, send_robust=send_robust, **kwargs)
+
+    def send_event_in_bus_consumer(
```
rgraber, line 197: I find this name really confusing and possibly too specific. Theoretically one could use this outside the context of the event bus, and we're not in the event-bus folder. What about either moving this to the event bus folder or changing it to something like
```suggestion
    def send_event_with_custom_metadata(
```
If we move it to the event bus folder, I would vote we call it `send_event_in_consumer` since no where else do we just call it the "bus."

openedx_events/data.py
robrap: Correct. It doesn't do that for us. Not sure if there is another way, but `id="mock-id"` works without the validator.

openedx_events/tooling.py
robrap: @ormsbee: FYI: Are you ok with this to avoid a breaking change? I know you wish time was required, and if we don't do that now, we probably will never do that.

openedx_events/tooling.py
robrap: Done.

openedx_events/tooling.py
robrap: Oops. I missed that you used the word "custom". Let me know if you think my change is ok, or if you are wed to your name.

openedx_events/data.py
rgraber: Sigh

openedx_events/tooling.py
```diff
@@ -185,6 +157,86 @@ def validate_sender():
...
+
+        Exceptions raised:
+            SenderValidationError: raised when there's a mismatch between
+            arguments passed to this method and arguments used to initialize
+            the event.
+        """
+        metadata = self.generate_signal_metadata(time=time)
+        return self._send_event_with_metadata(metadata=metadata, send_robust=send_robust, **kwargs)
+
+    def send_event_with_provided_metadata(
```
rgraber, line 203: [nit] I prefer `custom` because when I see `provided` it makes me think that the data is somehow provided externally (as opposed to being set by the caller). Preference is not enough to block the review, however.

rgraber (approved):
Nit about naming, but otherwise LGTM

openedx_events/tooling.py
robrap: Note: I plan to merge this. We can always refactor later. Clearly, the sooner the better, if we plan to break things.

PR #177, opened 2023-02-01
test: Check that there are no data/metadata key collisions

timmc-edx:
This will ensure that no signals are added that conflict with metadata fields, or vice versa. Ensuring that there are no collisions will become important once the event bus consumer starts sending metadata to receivers.

**Merge checklist:**
- [ ] Version bumped
- [ ] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/openedx_events/tests/test_data.py
+++ b/openedx_events/tests/test_data.py
@@ -1,0 +1,39 @@
+"""
+Tests for data.py.
+"""
+import attr
+import pytest
+from django.test import TestCase
+
+from openedx_events.data import EventsMetadata
+from openedx_events.tests.utils import load_all_signals
+from openedx_events.tooling import OpenEdxPublicSignal
+
+
+class EventsMetadataTestCache(TestCase):
+    """
+    Tests for EventsMetadata.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        load_all_signals()
+
+    def test_key_non_collision(self):
+        """
+        Check whether any events have a data key that collides with a metadata key.
+
+        This is important because both the data dict and the metadata dict get
+        splatted into the signal receiver, so we need to make sure none of the keys
+        collide.
+        """
+        metadata_fields = set(f.name for f in attr.fields(EventsMetadata))
+        for signal in OpenEdxPublicSignal.all_events():
+            for signal_root_key in signal.init_data.keys():
+                if signal_root_key in metadata_fields:
+                    pytest.fail(
+                        f"Signal {signal} has a root data key called {signal_root_key!r}, "
+                        "but this collides with an EventMetadata field. This signal will "
+                        "need to use a different key at the root of its data dictionary."
+                    )
--- a/openedx_events/event_bus/avro/tests/test_avro.py
+++ b/openedx_events/event_bus/avro/tests/test_avro.py
@@ -4,10 +4,6 @@ from unittest import TestCase
 
 from opaque_keys.edx.keys import CourseKey, UsageKey
 
-# Each new folder with signals must be manually imported in order for the signals to be cached
-#   and used in the unit tests. Using 'disable=reimported' with pylint will work,
-#   because we just use the cached signal list.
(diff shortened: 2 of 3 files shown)


timmc-edx:
For reference:

- https://docs.python.org/3/library/importlib.html#importlib.import_module
- https://docs.python.org/3.8/library/pkgutil.html#pkgutil.walk_packages

openedx_events/tests/test_data.py
```diff
@@ -0,0 +1,39 @@
...
+
+        This is important because both the data dict and the metadata dict get
+        splatted into the signal receiver, so we need to make sure none of the keys
+        collide.
+        """
+        metadata_fields = set(f.name for f in attr.fields(EventsMetadata))
+        for signal in OpenEdxPublicSignal.all_events():
+            for signal_root_key in signal.init_data.keys():
+                if signal_root_key in metadata_fields:
+                    pytest.fail(
```
timmc-edx, line 35: I'll probably just put a nocover comment on this.

openedx_events/tests/test_data.py
```diff
@@ -0,0 +1,39 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+        load_all_signals()
+
+    def test_key_non_collision(self):
+        """
+        Check whether any events have a data key that collides with a metadata key.
+
+        This is important because both the data dict and the metadata dict get
+        splatted into the signal receiver, so we need to make sure none of the keys
```
robrap, line 28: I don't understand. When does this happen? I thought metadata was kept as a dict. You are saying that all of its keys are copied from the dict into separate arguments?

timmc-edx: Aha! I've misread tooling.py. I had thought it happens in `send_event`, but you're right, it gets packed back up into a key called `metadata`. It actually happens in the call to `send_event_with_custom_metadata`: https://github.com/openedx/openedx-events/blob/bab0b3930cf7761db146e24be819bb90249223fc/openedx_events/tooling.py#L203

Here's where the call would happen, in a pending PR: https://github.com/openedx/event-bus-kafka/blob/6477f20e2680bb0bac972bbc987cf049c31e2e09/edx_event_bus_kafka/internal/consumer.py#L365

Maybe we can change that to just accept a single `metadata: EventsMetadata`? Breaking change, but we're not using it yet. cc: @rgraber

robrap: 1. We intentionally don’t include the event type argument, which is not a big deal, but a point of earlier discussion.
2. I imagine the code to create the EventsMetedata object would look very similar to this, so it seemed ok to encapsulate it.
3. What happens when we add new metadata? Would the current code make it more or less clear that something needs to change? I think I have a unit test around this, but if not, we should.

I did ponder both ways and chose the current implementation. If you feel strongly it should change, can you document the benefits? I don’t feel very strongly, except that it is done and could have gone either way.

timmc-edx: Yeah, I should probably go and look at the discussion on that PR before having too many opinions on it. :-) Anyway, this is the place where collisions would happen, and which this PR is relevant to. (Maybe also check that signals don't use `metadata` or (implausibly) `send_robust`? Or maybe not.)

robrap: Hah. Now I get it. It wasn’t clicking. I didn’t realize how the data objects were sent. Maybe it is safer to use EventsMetadata after all. We might want a factory-like method to create the EventsMetadata, but I agree that we don’t want to be concerned with conflicts.

timmc-edx:
Filed https://github.com/openedx/openedx-events/issues/178 to change the API and submitted https://github.com/openedx/openedx-events/pull/179 to implement it. If that is merged, then this can just be reduced to the auto-detection of signals.py files without any of the new validation.

timmc-edx:
As we're no longer at risk of collisions, I've reworked this into a smaller PR that just removes reliance on hardcoded imports: https://github.com/openedx/openedx-events/pull/181

PR #179, opened 2023-02-02
feat!: Only accept metadata object in `send_event_with_custom_metadata`

timmc-edx:
This changes `send_event_with_custom_metadata` to accept a single `metadata` object rather than all the fields that go into one.

It turns out that in practice, the event-bus-kafka consumer already needs to build an EventsMetadata object or something equivalent to it as it reconstructs the metadata from Kafka message headers, since that's where the parsing and validation logic already is. It then has to splat the object back out when calling this method. This leads to a possible collision risk in the method's kwargs, but it's also just awkward. It is likely that the same issue would arise in any other consumer.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed


openedx_events/tooling.py
```diff
@@ -185,8 +185,7 @@ def send_event(self, send_robust=True, time=None, **kwargs):
             [(<function callback at 0x7f2ce638ef70>, 'callback response')]
 
         Keyword arguments:
```
rgraber, line 187: [nit][not-blocking]
I find it confusing to have the arguments split up like this. Can we move this up with the other arguments?

rgraber (approved):
LGTM. Good catch.

openedx_events/tooling.py
timmc-edx: Sure! I was mimicking what I saw elsewhere in the file, but now that I check, that doesn't match our style guide. I'll fix it up.

PR #181, opened 2023-02-03
test: Automatically discover new signals.py files in tests

timmc-edx:
This will ensure that when new signals are added in new domains, they'll be picked up by the tests. There's still a hardcoded list, but it's just there to ensure there aren't silent regressions.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/openedx_events/tests/utils.py
+++ b/openedx_events/tests/utils.py
@@ -1,6 +1,9 @@
 """
 Utils used by Open edX event tests.
 """
+import pkgutil
+from importlib import import_module
+
 from openedx_events.tooling import OpenEdxPublicSignal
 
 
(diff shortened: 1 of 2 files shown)


rgraber (approved):
LGTM

openedx_events/tests/utils.py
```diff
@@ -124,3 +127,32 @@ def start_events_isolation(cls):
...
+
+    Loads all non-test signals.py modules.
+    """
+    found = set()
+
+    root = import_module('openedx_events')
+    for m in pkgutil.walk_packages(root.__path__, root.__name__ + '.'):
+        module_name = m.name
+        if 'tests' in module_name.split('.') or '.test_' in module_name:
+            continue
```
timmc-edx, line 144: I don't actually care about coverage here, but it's _weird_ that this line is allegedly not covered. There should be about 10 test modules that hit this line during the walk, and I see them locally if I put a print statement here!

nedbat: This is because Python has optimized away that line: https://github.com/nedbat/coveragepy/issues/198

PR #187, opened 2023-02-23
refactor!: move load_all_events to more visible location

rgraber:
**Description:**
Move load_all_events out of the test-only module to make it clear that it can be used by external apps that use openedx_events.

It seems a little bit silly to make this a breaking change but technically it would break anything that tried to use load_all_events. Nothing *should* be using it but we can't guarantee that.

**ISSUE:** https://github.com/openedx/event-bus-kafka/issues/135

**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/openedx_events/tooling.py
+++ b/openedx_events/tooling.py
@@ -1,7 +1,9 @@
 """
 Tooling necessary to use Open edX events.
 """
+import pkgutil
 import warnings
+from importlib import import_module
 from logging import getLogger
 
 from django.conf import settings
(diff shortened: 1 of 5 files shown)


openedx_events/tooling.py
```diff
@@ -252,3 +254,31 @@ def allow_send_event_failure(self):
...
+    # Check that the auto-discovered list matches the known modules.
+    # This is just here to ensure that the auto-discovery is working
+    # properly and doesn't start to silently fail.
+    #
+    # If this assertion fails because a module has been added, renamed,
+    # or deleted, please update the hardcoded list.
+    assert found == {
+        'openedx_events.content_authoring.signals',
+        'openedx_events.learning.signals',
+    }
```
dianakhuang, line 284: Part of me is a little dubious about hardcoding this list and auto-discoverying it, but I guess this is a good canary for when new signals are added.

dianakhuang (approved):

dianakhuang (approved):
I don't think the test is as bad as you think it is, even if I also feel scared of messing about internals that much.

Issue #78, opened 2022-07-12
[Discovery] Handle multiple lifecycle events on single event bus topic


robrap:
This is a bit of a discovery ticket.

Acceptance Criteria:

- [x] ADR for handling of multiple lifecycle events (created, updated, deleted) on a single topic.
- [ ] Update how-to as appropriate.
- [x] Create follow-up ticket (if needed) for any work that would enable onboarding new multi-event type topics sooner rather than later.

Notes:

- Confluent recommends keeping multiple related event types on the same topic for ordering purposes:
  - See https://www.confluent.io/blog/put-several-event-types-kafka-topic/
  - And more recent article: https://www.confluent.io/blog/multiple-event-types-in-the-same-kafka-topic/

Questions:

- Will we follow Confluent’s recommendation to use a single topic for multiple related events?
  - How will we get a partition key from our current event definition?
  - How will we handle the schema used for the topic to allow for the multiple event types?
    - Note: the second article (https://www.confluent.io/blog/multiple-event-types-in-the-same-kafka-topic/) offers two options with different pros/cons.
    - [Robert] I'm thinking the Union solution is probably more complicated than what we need. However, I'd like to understand how schema evolution works with the other solution.
- Will this solution work for other event bus technologies other than Kafka?
- What are good naming conventions for the `event_type`, event definition (signal), data class, etc. for the proposed solution?

robrap:
Moved from original private ticket: https://2u-internal.atlassian.net/browse/ARCHBOM-2008

robrap:
Move some of these notes into the doc ticket, and closing this.

robrap:
Note: This was closed without completing, and will remain closed (for now). However, I just updated the how-to to refer to this closed ticket: https://openedx.atlassian.net/wiki/spaces/AC/pages/3508699151/How+to+start+using+the+Event+Bus#Event-design.

Other considerations for ordering will include Redis Streams capabilities, which don't have the same partiton ordering guarantees as Kafka.

Here are docs that I moved from the how-to back to this ticket here:

> Confluent, a hosted solution for Kafka, recommends keeping multiple related event types on the same topic for ordering purposes. If going for the approach of multiple lifecycle events in one topic, you will want to consider the partition key carefully. See [Publishing a signal](https://openedx.atlassian.net/wiki/spaces/AC/pages/3508699151/How+to+start+using+the+Event+Bus#Publishing-a-signal) for more.
>
[...]

PR #212, opened 2023-05-03
feat: add namespace to schemas

rgraber:
**Description:**
Add namespace to event schemas. This will allow us to use the TopicRecordName strategy when registering schemas (otherwise, all the record names would just be "CloudEvent" and there would be name conflicts). This should be able to be deployed the producer and consumer in any order.

This will make the schema for a topic look like
```
{
 "doc": "Avro Event Format for CloudEvents created with openedx_events/schema",
  "fields": [<no change>],
  "name": "CloudEvent",
  "namespace": "org.openedx.content_authoring.course.catalog_info.changed.v1",
  "type": "record"
}
```

**ISSUE:**
https://github.com/openedx/openedx-events/issues/78

**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/openedx_events/event_bus/avro/tests/test_schema.py
+++ b/openedx_events/event_bus/avro/tests/test_schema.py
@@ -34,6 +34,7 @@ class TestSchemaGeneration(TestCase):
             "type": "record",
             "name": "CloudEvent",
             "doc": "Avro Event Format for CloudEvents created with openedx_events/schema",
+            'namespace': 'simple.signal',
             "fields": [
                 {"name": "event_data", "type":
                     {"name": "SimpleAttrs", "type": "record", "fields": [
@@ -58,6 +59,7 @@ class TestSchemaGeneration(TestCase):
             "type": "record",
             "name": "CloudEvent",
             "doc": "Avro Event Format for CloudEvents created with openedx_events/schema",
+            'namespace': 'simple.signal',
             "fields": [
                 {"name": "test_data", "type":
                     {"name": "EventData", "type": "record", "fields": [
@@ -97,6 +99,7 @@ class TestSchemaGeneration(TestCase):
             "type": "record",
             "name": "CloudEvent",
             "doc": "Avro Event Format for CloudEvents created with openedx_events/schema",
+            'namespace': 'simple.signal',
             "fields": [
                 {"name": "top_level_key_0", "type":
                     {
@@ -128,6 +131,7 @@ class TestSchemaGeneration(TestCase):
             "type": "record",
             "name": "CloudEvent",
             "doc": "Avro Event Format for CloudEvents created with openedx_events/schema",
+            'namespace': 'simple.signal',
             "fields": [
                 {"name": "test_data", "type": "string"}
             ],
@@ -141,6 +145,7 @@ class TestSchemaGeneration(TestCase):
             "type": "record",
             "name": "CloudEvent",
             "doc": "Avro Event Format for CloudEvents created with openedx_events/schema",
+            'namespace': 'simple.signal',
             "fields": [
                 {"name": "test_data", "type": {
                     "name": "NestedNonAttrs",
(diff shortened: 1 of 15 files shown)


mariajgrimaldi (approved):
LGTM :). Can you add an example of how this would look like? So it's documented

rgraber:
Ack, I should have left this in draft while I was still testing, sorry! I'm just making sure it can be deployed independently on producers and consumers without issue

robrap:
@rgraber: Does this break CloudEvent requirements? See https://github.com/cloudevents/spec/blob/v1.0.1/avro-format.md#24-definition. If it does, we'll need determine if we move forward anyway and just note this somewhere. Note that @ormsbee selected CloudEvent because it seemed useful (he can say more), at the same time, I don't think it matters for anything at this point. Depending on your answer, we can discuss with Dave.

rgraber:
> @rgraber: Does this break CloudEvent requirements? See https://github.com/cloudevents/spec/blob/v1.0.1/avro-format.md#24-definition. If it does, we'll need determine if we move forward anyway and just note this somewhere. Note that @ormsbee selected CloudEvent because it seemed useful (he can say more), at the same time, I don't think it matters for anything at this point. Depending on your answer, we can discuss with Dave.

It's unclear to me if that exact namespace is required. If anything, this would be a lateral move since we didn't have any namespace there to begin with so I don't know if we need to answer that now.

rgraber:
@mariajgrimaldi this is ready for review, sorry about the mixup earlier

openedx_events/event_bus/avro/tests/test_schema.py
```diff
@@ -34,6 +34,7 @@ def test_simple_schema_generation(self):
             "type": "record",
             "name": "CloudEvent",
             "doc": "Avro Event Format for CloudEvents created with openedx_events/schema",
+            'namespace': 'simple.signal',
```
robrap, line 37: Nit: All this repetition makes me think a helper for generating this data might clean-up tests.

robrap (approved):
Looks fine, but please update the commit comment (and maybe PR description) with more details around why this is being added, and potentially pointing to docs regarding its need. Thank you!

robrap:
@rgraber: That's fine. It was more about the commit message being empty. If there isn't a good doc to point to, forget about that.

robrap:
@rgraber: In Slack, but possibly not here, you had written:
> It’s something we should do regardless of how we go about putting multiple events on one topic because it distinguishes the names of schemas between different events (since they’re all called “CloudEvent”).

Since we probably aren't using TopicRecordName strategy (please confirm), then shouldn't we either close this PR, or better document why we are doing this? I think your Slack comment is why you think it is important, right? Thanks.

rgraber:
> @rgraber: In Slack, but possibly not here, you had written:
> 
> > It’s something we should do regardless of how we go about putting multiple events on one topic because it distinguishes the names of schemas between different events (since they’re all called “CloudEvent”).
> 
> Since we probably aren't using TopicRecordName strategy (please confirm), then shouldn't we either close this PR, or better document why we are doing this? I think your Slack comment is why you think it is important, right? Thanks.

Ah, I see looking back in Slack that I was not very clear.
We probably will still use the TopicRecordName strategy but we will be setting it on the client rather than the broker. This we can do regardless of cluster type. Thus, we will still need this PR to ensure that the schema for each signal type on a topic gets registered separately.
I believe we will need this PR even if we change our minds about using the TopicRecordName strategy vs avro unions because Avro unions refer to other schemas by name, so we'll need the names to be distinct.

rgraber:
Am I still good to merge?

rgraber:
I'm going to go ahead and merge so we can unblock the xblock work. Will put the rationale in the extended commit message.

Issue #218, opened 2023-05-12
Remove signal parameter from consumer


rgraber:
The signal parameter of the consumer init method and the management command have been deprecated and should be removed. It may be necessary to make them optional first, release that, then remove them as a new major version. 

A/C:

- [x] Make optional in openedx-events (and event-bus-redis if it isn't already. it's already optional in event-bus-kafka)
- [x] Remove parameter from all active consumers
   - [x] make sure to test on stage before removing in prod 
- [x] Remove parameter from openedx-events
- [x] Remove parameter from EBR and EBK

PR #219, opened 2023-05-12
feat: make signal parameter optional in preparation for removal

rgraber:
**Description:**
Makes the `signal` parameter of the consume_events command optional in preparation for complete removal. The signal will now be determined from the message headers.

**ISSUE:**
https://github.com/openedx/openedx-events/issues/218

**Dependencies:**
If we're being nice, should probably not be merged before https://github.com/openedx/event-bus-redis/pull/22, although the event bus redis is not production-ready yet anyway.
**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.


openedx_events/event_bus/__init__.py
```diff
@@ -157,7 +157,9 @@ def consume_indefinitely(self) -> None:
 #   an instance of EventBusConsumer, calls to the consumer will be ignored with a warning at startup.
 
 
-def make_single_consumer(*, topic: str, group_id: str, signal: OpenEdxPublicSignal, **kwargs) -> EventBusConsumer:
+def make_single_consumer(*, topic: str, group_id: str,
+                         signal: OpenEdxPublicSignal = None,  # pylint: disable=unused-argument
+                         **kwargs) -> EventBusConsumer:
     """
     Construct a consumer for a given topic, group, and signal.
```
robrap, line 164: 1. Do you want to drop `signal` from here now?
2. Should we add a sentence stating that the signals will be determined by the message header, or do we not even need to mention signals in the future?

openedx_events/event_bus/__init__.py
```diff
@@ -166,12 +168,11 @@ def make_single_consumer(*, topic: str, group_id: str, signal: OpenEdxPublicSign
     Arguments:
         topic: The event bus topic to consume from (without any environmental prefix)
         group_id: The consumer group to participate in
-        signal: Send consumed, decoded events to the receivers of this signal
+        signal: DEPRECATED signals determined by message headers
```
robrap, line 171: If you add the notion of `signals determined by message headers` to a more permanent comment, then maybe the following would be more clear?
```suggestion
        signal: (DEPRECATED) this argument will be ignored.
```

openedx_events/management/commands/consume_events.py
```diff
@@ -54,8 +51,8 @@ def add_arguments(self, parser):
         parser.add_argument(
             '-s', '--signal',
             nargs=1,
-            required=True,
-            help='Type of signal to emit from consumed messages.'
+            required=False,
+            help='DEPRECATED signal will be determined by message headers'
```
robrap, line 55: See related comment and do what you think is best.

robrap (approved):
Thoughts around comments, but non-blocking.

openedx_events/event_bus/__init__.py
rgraber: 1. Going to keep signal here for now since this is part of the public API. I don't know of anyone who uses it but I've broken things before by unexpectedly changing a public API method
2. I think we should just not mention it

Issue #210, opened 2023-05-01
Enable producing to event bus via settings


robrap:
At this time, producing a signal to the event bus requires code. See the [event bus how-to for details](https://openedx.atlassian.net/wiki/spaces/AC/pages/3508699151/How+to+start+using+the+Event+Bus#Producing-a-signal). 

This is problematic because:
1. Nearly all openedx-events signals should be able to be produced to the event bus, and different operators may have different needs.
2. Related, for named releases, operators should be able to produce new events without changing code or waiting for the next named release.

This issue proposes to make the switch to configuring this via settings, rather than code.

Some thoughts/questions about the settings:
1. It should be possible to toggle on/off an event as we do in code. I imagine that we'd want to be able to do this without having to add/delete all config for a topic, and could instead set a flag.
2. Does the setting need to support producing the same event to multiple topics? Maybe not? I can't remember where we landed for migrations, and maybe we could add this later while keeping a simpler one-topic setup by default?
3. The settings should support multiple events going to the same topic. See this lifecycle event issue for details on where we landed on this: https://github.com/openedx/openedx-events/issues/78.
4. It would be great if services could include default settings for signals that are expected to be produced. These may be toggled off, or incomplete, but could be used as documentation of what signals should be configured. (Or, someone can provide an alternative solution to this particular issue.)
5. This may be best covered in an entirely separate ticket, but it would be nice if we could toggle some sort of audit that would help estimate the data that would be sent to the event bus if a event were sent there. This could be used for making cost estimates for existing events.
6. Reminder: replace existing event producing code with this config when it is available.

bmtcril:
@navinkarkera is going to take this on next and write an ADR up front for the proposed solution. I can't assign it to him, though. :)

robrap:
@rgraber: Adding a reminder for https://github.com/openedx/openedx-events/pull/272/files#r1347321249, which is to not have the configs reading fail hard for a single part that is misconfigured.
- [ ] Allow partial loading of config (with proper error handling to hopefully catch what is broken). 

Questions: 
- Do we want to alert on this? 
- Will we still raise an error (at the end)? Or rely on logs?

rgraber:
I'm going to create a separate ticket to deal with that. Seems like it requires more discussion.

rgraber:
Closing this and moving the other question to https://github.com/openedx/openedx-events/issues/280

robrap:
[...]

PR #224, opened 2023-05-23
docs: adr for enabling producing to event bus via settings

navinkarkera:
**Description:** ADR for enabling producing to event bus via settings

**ISSUE:** https://github.com/openedx/openedx-events/issues/210

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/docs/decisions/0012-producing-to-event-bus-via-settings.rst
+++ b/docs/decisions/0012-producing-to-event-bus-via-settings.rst
@@ -1,0 +1,90 @@
+11. Enable producing to event bus via settings
+##############################################
+
+Status
+******
+
+**Accepted** 2023-06-05
+
+Context
+*******
+
+The initial implementation of the event bus allowed only a single event type to be published to a `topic`_/`stream`_, with details like topic/stream name, consumer group name, and consumer name configured via code. We weren't sure what the API would finally look like, and whether the event bus config would live as part of the definition of the event, so we just went with explicit code that we could iterate upon.
+
+The current implementation of openedx-events does not actually push any events to the underlying implementations like `edx-event-bus-kafka`_ and `edx-event-bus-redis`_. The event-producing application is expected to create a signal handler (since openedx-events subclasses Django signals) to catch the event and push it into the event bus. Some examples of the handlers: `handlers example`_.
+
+.. _handlers example: https://github.com/openedx/edx-platform/blob/27b8d2f68d5dfaf84755e7d7f8dccc97ce3be509/cms/djangoapps/contentstore/signals/handlers.py#L162-L210
+.. _edx-event-bus-kafka: https://github.com/openedx/event-bus-kafka
+.. _edx-event-bus-redis: https://github.com/openedx/event-bus-redis
+.. _topic: https://developer.confluent.io/learn-kafka/apache-kafka/topics/#kafka-topics
+.. _stream: https://redis.io/docs/data-types/streams/
+
+This ADR aims to propose a solution for configuring the details, like the topic name, consumer group name, etc. via Django settings as well as pushing events to the event bus without requiring the producing application to write additional handlers.
+
+
+Decision
+********
+
+Create a generic signal handler to push events to the event bus. This handler should be attached to or connected to the signals that are enabled in Django settings. The configuration format will be as shown below:
+
+.. code-block:: python
+
+   # .. setting_name: EVENT_BUS_PRODUCER_CONFIG
+   # .. setting_default: {}
+   # .. setting_description: Dictionary of event_types to lists of dictionaries for topic related configuration.
+   #    Each topic configuration dictionary contains a flag called `enabled` denoting whether the event will be
+   #    published to the topic, topic/stream name called `topic` where the event will be pushed to,
+   #    `event_key_field` which is a period-delimited string path to event data field to use as event key.
+   #    Note: The topic names should not include environment prefix as it will be dynamically added based on
+   #    EVENT_BUS_TOPIC_PREFIX setting.
+   EVENT_BUS_PRODUCER_CONFIG = {
+       'org.openedx.content_authoring.xblock.published.v1': [
+           {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
+           {'topic': 'content-authoring-xblock-published', 'event_key_field': 'xblock_info.usage_key', 'enabled': False},
+       ],
+       'org.openedx.content_authoring.xblock.deleted.v1': [
+           {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
+       ],
+   }
+
+This configuration will be read in openedx_events ``apps.OpenedxEventsConfig(AppConfig).ready`` method and a generic signal handler will be connected to the event_types (keys) listed in the configuration after validating its format.
+
+.. code-block:: python
+
+   def ready(self):
+       load_all_signals()
+       config = read_config()
+       # validate the config dictionary and raise errors if any
+       validate_config()
+       for configured_signal in config:
+           connect_or_disconnect_handlers(configured_signal)
+
+The generic handler will again read the configuration and get details for event or signal triggered and push it to the event bus.
+
+.. code-block:: python
+
+   def general_signal_handler(sender, signal, **kwargs):
+       config = read_config(signal)
+       for topic in topics:
+           push_event_to_implementation(signal, config)
+
+The `push_event_to_implementation` function is the same code currently used in the `hard-coded handlers`_, the only difference is that it will read the details from the configuration.
+
+Finally, replace the `handlers`_ in edx-platform with a default configuration that will configure and enable the same events as the handlers.
+
+.. _handlers: https://github.com/openedx/edx-platform/blob/27b8d2f68d5dfaf84755e7d7f8dccc97ce3be509/cms/djangoapps/contentstore/signals/handlers.py#L162-L210
+.. _hard-coded handlers: https://github.com/openedx/edx-platform/blob/27b8d2f68d5dfaf84755e7d7f8dccc97ce3be509/cms/djangoapps/contentstore/signals/handlers.py#L167-L171
+
+Consequences
+************
+
+* Applications producing to event_bus can push events without requiring any additional code.
+* Users can push multiple related event_types to the same topic, which will allow them to run a single consumer and process the events in the correct order.
+* The configuration is a dictionary, which makes it flexible but difficult to enforce schemas and maintain them.
+* Producing applications similar to consuming applications will be required to add ``openedx_events`` as an application in their ``INSTALLED_APPS`` settings.
+
+Rejected Alternatives
+*********************
+
+* Implementing configurable handlers in the host applications, rather than in the shared library, will require repeating the code in each host application.
+* Following the current way of using fixed handlers will restricts users ability to combine events in a single topic based on their preference, as well as stopping users from producing additional defined events to the event bus without code changes.
--- a/docs/decisions/index.rst
+++ b/docs/decisions/index.rst
@@ -16,3 +16,4 @@ Architectural Decision Records (ADRs)
    0009-course-catalog-info-changed-design
    0010-multiple-event-types-per-topic
    0011-depending-on-multiple-event-bus-implementations
+   0012-producing-to-event-bus-via-settings


openedx-webhooks:
Thanks for the pull request, @navinkarkera!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,106 @@
...
+implementations like `edx-event-bus-kafka`_ and `edx-event-bus-redis`_. The
+event-producing application is expected to create a signal handler (since
+openedx-events subclasses Django signals) to catch the event and push it into
+the event bus. Some examples of the handlers can be found `here`_.
+
+.. _here: https://github.com/openedx/edx-platform/blob/27b8d2f68d5dfaf84755e7d7f8dccc97ce3be509/cms/djangoapps/contentstore/signals/handlers.py#L162-L210
+.. _edx-event-bus-kafka: https://github.com/openedx/event-bus-kafka
+.. _edx-event-bus-redis: https://github.com/openedx/event-bus-redis
+
+This ADR aims to propose a solution for configuring the details, like the topic.
```
timmc-edx, line 25: [nit] stray character
```suggestion
This ADR aims to propose a solution for configuring the details, like the topic
```

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Status
+******
+
+**Provisional** 2023-05-23
+
+Context
+*******
+
+Since the initial implementation of the event bus allowed only a single event
```
timmc-edx, line 12: OEP-19 requests that hard line breaks be avoided in paragraphs in rST files: https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0019-bp-developer-documentation.html#best-practices

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,106 @@
...
+the event bus without requiring the producing application to write additional
+handlers.
+
+
+Decision
+********
+
+Create a generic signal handler to push events to the event bus. This handler
+should be attached to or connected to the signals that are enabled in Django settings.
+of the application. The configuration format will be as shown below:
```
timmc-edx, line 36: Possibly a stray sentence fragment here?

```suggestion
The configuration format will be as shown below:
```

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,106 @@
...
+           'event_key_field': 'xblock_info.usage_key',
+           'event_data_key': 'xblock_info',
+           'enabled': True,
+       },
+       'org.openedx.content_authoring.xblock.published.v1': {
+           'topics': ['xblock-status', 'xblock-published'],
+           'event_key_field': 'xblock_info.usage_key',
+           'event_data_key': 'xblock_info',
+           'enabled': True,
+       }
```
timmc-edx, line 52: We should probably have a trailing comma here even though this is example code.
```suggestion
       },
```

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,106 @@
...
+   }
+
+This configuration will be read in openedx_events ``apps.OpenedxEventsConfig(AppConfig).ready``
+method and a generic signal handler will be connected to the event_types (keys)
+listed in the configuration after validating its format.
+
+.. code-block:: python
```
timmc-edx, line 59: I don't know that there's much benefit to including implementation code in the ADR, as the actual code will diverge over time. It's probably sufficient to say that `openedx_events` will implement a Django `AppConfig.ready()` method to load the config and register signal handlers.

It might actually make sense to just have one PR that adds implementation and ADR at the same time, which we've done pretty often. I think the implementation approach should be uncontroversial enough that it would make sense to just go ahead and do that.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Create a generic signal handler to push events to the event bus. This handler
+should be attached to or connected to the signals that are enabled in Django settings.
+of the application. The configuration format will be as shown below:
+
+.. code-block:: python
+
+   OPEN_EDX_EVENTS_CONFIG = {
+       'org.openedx.content_authoring.xblock.deleted.v1': {
+           'topics': 'xblock-status',
```
timmc-edx, line 42: Even if we end up mostly having single-topic events, I have a preference for always requiring a list here. Interested to hear how others feel about that.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,106 @@
...
+  additional code.
+* The configuration is a dictionary, which makes it flexible but difficult to
+  enforce schemas and maintain them.
+* Producing as well as consuming applications will be required to add
+  ``openedx_events`` as an application in their ``INSTALLED_APPS`` settings.
+
+Rejected Alternatives
+*********************
+
+None
```
timmc-edx, line 106: It's probably worth a sentence or two describing the downsides of the current approach, which is an alternative that we're rejecting.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,106 @@
...
+Since the initial implementation of the event bus allowed only a single event
+type to be published to a topic, details like topic name, consumer group name,
+and consumer name were configured via code. The current implementation of
+openedx-events does not actually push any events to the underlying
+implementations like `edx-event-bus-kafka`_ and `edx-event-bus-redis`_. The
+event-producing application is expected to create a signal handler (since
+openedx-events subclasses Django signals) to catch the event and push it into
+the event bus. Some examples of the handlers can be found `here`_.
+
+.. _here: https://github.com/openedx/edx-platform/blob/27b8d2f68d5dfaf84755e7d7f8dccc97ce3be509/cms/djangoapps/contentstore/signals/handlers.py#L162-L210
```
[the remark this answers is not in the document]
navinkarkera: Updated.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: @timmc-edx Makes sense, I wanted to include example of code to make it easier to understand. Updated it to just pseudo code now.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: +1

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: Done.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,74 @@
...
+* Applications producing to event_bus can push events without requiring any additional code.
+* Users can push related events to the same topic, which will allow them to run a single consumer and process the events in the correct order.
+* The configuration is a dictionary, which makes it flexible but difficult to enforce schemas and maintain them.
+* Producing as well as consuming applications will be required to add ``openedx_events`` as an application in their ``INSTALLED_APPS`` settings.
+
+Rejected Alternatives
+*********************
+
+* Implementing configurable handlers in the host applications will require repeating the code in each host application.
+* Following the current way of using fixed handlers will force users to create multiple consumer processes for consuming multiple events, and hence the order of execution is not guaranteed across events.
```
timmc-edx, line 74: I'm not sure how this relates to the configurable producer part.

navinkarkera: Having the possibility of configuring the topic name while producing events allows user to consume related events using a single consumer.

mphilbrick211:
@mariajgrimaldi would you be able to review/merge this?

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,76 @@
+11. Enable producing to event bus via settings
+##############################################
+
+Status
+******
+
+**Provisional** 2023-05-23
```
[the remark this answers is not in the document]
navinkarkera: @robrap Some of the ADRs still have their status set to Provisional, so I was not sure when to set it as accepted. I'll update this and the others to accepted if it makes sense.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
Ian2012: cc @mariajgrimaldi

docs/decisions/0011-producing-to-event-bus-via-settings.rst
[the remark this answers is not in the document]
robrap: I think it is fine for you to do whatever you wish.

The following is from [OEP-19](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0019-bp-developer-documentation.html#adrs):
> If an ADR has Draft status and the PR is under review, you can either use the intended final status (e.g. Provisional, Accepted, etc.), or you can clarify both the current and intended status using something like the following: "Draft (=> Provisional)". Either of these options is especially useful if the merged status is not intended to be Accepted.

I think the main point of Provisional is when we are making a decision, but we are uncertain about it, especially if there will be a big gap between the decision and implementation. In this case, I think you can save yourself some time going straight to Accepted, but anything is fine.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: @robrap Makes sense, how about updating it to:

* Following the current way of using fixed handlers will restricts users ability to combine events in a single topic based on their preference.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,76 @@
...
+* Applications producing to event_bus can push events without requiring any additional code.
+* Users can push related events to the same topic, which will allow them to run a single consumer and process the events in the correct order.
+* The configuration is a dictionary, which makes it flexible but difficult to enforce schemas and maintain them.
+* Producing as well as consuming applications will be required to add ``openedx_events`` as an application in their ``INSTALLED_APPS`` settings.
+
+Rejected Alternatives
+*********************
+
+* Implementing configurable handlers in the host applications will require repeating the code in each host application.
+* Following the current way of using fixed handlers will force users to create multiple consumer processes for consuming multiple events, and hence the order of execution is not guaranteed across events.
```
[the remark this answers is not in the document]
navinkarkera: @robrap Just to make sure that I understand, by no simple way you mean except for writing code which pushes events to the event bus right?

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,76 @@
...
+The generic handler will again read the configuration and get details for event or signal triggered and push it to the event bus.
+
+.. code-block:: python
+
+   def general_signal_handler(sender, signal, **kwargs):
+       config = read_config(signal)
+       for topic in topics:
+           push_event_to_implementation(signal, config)
+
+Consequences
```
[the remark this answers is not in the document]
navinkarkera: We can probably update [OpenEdxPublicSignal](https://github.com/open-craft/openedx-events/blob/4db956585a831d2f40605b9480967fff452d7d0a/openedx_events/content_authoring/signals.py#L23-L28) to include an optional argument called topics and then use the these to generate a default configuration for each signal. This default config can then be updated with the values from users config allowing users to only define the params that they want to change. It would also remove the need for users to define `event_key_field` and `event_data_key` in the config.

I'll update the document on Monday with better explanation of my plan. :)

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,76 @@
...
+************
+
+* Applications producing to event_bus can push events without requiring any additional code.
+* Users can push related events to the same topic, which will allow them to run a single consumer and process the events in the correct order.
+* The configuration is a dictionary, which makes it flexible but difficult to enforce schemas and maintain them.
+* Producing as well as consuming applications will be required to add ``openedx_events`` as an application in their ``INSTALLED_APPS`` settings.
```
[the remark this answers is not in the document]
navinkarkera: Updated, actually I wanted to say that producing applications will also need to add them similar to consuming applications.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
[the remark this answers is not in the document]
navinkarkera: Done.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Status
+******
+
+**Provisional** 2023-05-23
+
+Context
+*******
+
+Since the initial implementation of the event bus allowed only a single event type to be published to a topic, details like topic name, consumer group name, and consumer name were configured via code. The current implementation of openedx-events does not actually push any events to the underlying implementations like `edx-event-bus-kafka`_ and `edx-event-bus-redis`_. The event-producing application is expected to create a signal handler (since openedx-events subclasses Django signals) to catch the event and push it into the event bus. Some examples of the handlers: `handlers example`_.
```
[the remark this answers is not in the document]
navinkarkera: Updated.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,76 @@
...
+   def general_signal_handler(sender, signal, **kwargs):
+       config = read_config(signal)
+       for topic in topics:
+           push_event_to_implementation(signal, config)
+
+Consequences
+************
+
+* Applications producing to event_bus can push events without requiring any additional code.
+* Users can push related events to the same topic, which will allow them to run a single consumer and process the events in the correct order.
```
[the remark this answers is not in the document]
navinkarkera: Updated.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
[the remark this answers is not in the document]
navinkarkera: Updated, please have a look at line 26.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,76 @@
...
+Decision
+********
+
+Create a generic signal handler to push events to the event bus. This handler should be attached to or connected to the signals that are enabled in Django settings. The configuration format will be as shown below:
+
+.. code-block:: python
+
+   OPEN_EDX_EVENTS_CONFIG = {
+       'org.openedx.content_authoring.xblock.deleted.v1': {
+           'topics': 'xblock-status',
```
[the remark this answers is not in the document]
navinkarkera: Updated.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,92 @@
...
+.. _edx-event-bus-kafka: https://github.com/openedx/event-bus-kafka
+.. _edx-event-bus-redis: https://github.com/openedx/event-bus-redis
+
+This ADR aims to propose a solution for configuring the details, like the topic name, consumer group name, etc. via Django settings as well as pushing events to the event bus without requiring the producing application to write additional handlers.
+
+
+Decision
+********
+
+Update constructor of ``OpenEdxPublicSignal`` class to optionally accept ``topics`` and ``event_key_field`` and store them. These values can be then used as default while pushing them to event bus implementations. Add ``topics`` and ``event_key_field`` for existing signal definitions in openedx_events repository. Once this ADR is implemented, make ``topics`` and ``event_key_field`` required in ``OpenEdxPublicSignal`` constructor.
```
robrap, line 26: I'd love to also get the thoughts of others, but I don't think I'd push this all the way to the event definition. I was thinking about the defaults simply being in the `EVENT_BUS_PRODUCER_CONFIG` that is defined in each service's repository. Thoughts?

robrap (commented):
I'll have someone else join in on the discussion of where to define defaults. Thank you.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,92 @@
...
+Optionally, add a function to ``OpenEdxPublicSignal`` to return default configuration for documentating them.
+
+Consequences
+************
+
+* Applications producing to event_bus can push events without requiring any additional code.
+* Users can push multiple related event_types to the same topic, which will allow them to run a single consumer and process the events in the correct order.
+* The configuration is a dictionary, which makes it flexible but difficult to enforce schemas and maintain them.
+* Producing applications similar to consuming applications will be required to add ``openedx_events`` as an application in their ``INSTALLED_APPS`` settings.
+* Adding details like ``topic`` and ``event_key_field`` to signal definition will keep configuration compact if user does not want to override them.
```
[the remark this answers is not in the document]
navinkarkera: >I could maybe see event_key_field being part of the definition

Agreed, we can maybe keep `event_key_field` with the definition and remove `topics`. This can be defined via the configuration only and add defaults to each service as you suggested above.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: Makes sense. We could add currently used events in the defaults. For example, add [these events](https://github.com/openedx/edx-platform/blob/27b8d2f68d5dfaf84755e7d7f8dccc97ce3be509/cms/djangoapps/contentstore/signals/handlers.py#L162-L210) for edx-platform.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,92 @@
...
+   # .. setting_description: Dictionary specifying event_types (keys) and its configuration dictionary as values.
+   #   Each configuration dictionary contains a flag denoting whether the event will be enabled or disabled and
+   #   optionally details like topics/stream names where the event will be pushed to, event_key_field which is
+   #   and period delimited string path to event data field to use as event key.
+   #   Note: The topic names should not include environment prefix as it will be dynamically added based on
+   #   EVENT_BUS_TOPIC_PREFIX setting.
+   EVENT_BUS_PRODUCER_CONFIG = {
+       'org.openedx.content_authoring.xblock.deleted.v1': {
+           'event_key_field': 'xblock_info.usage_key', # override default event_key_field
+           'enabled': True,
```
timmc-edx, line 43: Given that this `enabled` boolean would be inside a dictionary structure, we wouldn't be able to (readily) override it separately from the rest of the `EVENT_BUS_PRODUCER_CONFIG`. I'm not sure how much value it adds over e.g. commenting out the unwanted entries.

(In contrast, disabling *all* event production could be done by setting the producer callable reference to an empty string, so that no implementation is loaded.)

docs/decisions/0011-producing-to-event-bus-via-settings.rst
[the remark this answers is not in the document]
timmc-edx: I think I'd prefer to keep both topic and event key with the service config. Centralizing the information to the event definition could allow us to avoid repetition in configs, but I don't think any event is going to be produced by more than one service in a deployment; in general, events should be produced by one service and then consumed by one or more others. The configs also don't have any environment-specific information in them, so we shouldn't even see repetition between stage vs. production configs—this config would be able to live in the service's own base settings.

Given the expected lack of repetition, I think having the topic and key field be closer to the point of production rather than farther (requiring library releases in order to change) would make config changes easy, without even the need for a default/override system.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: @timmc-edx I am not sure if I follow the issue here. Could you please explain with an example.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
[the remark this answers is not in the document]
navinkarkera: > events should be produced by one service and then consumed by one or more others.

@timmc-edx Makes absolute sense. Since we won't have any repetition as it is, we don't need to add these to the signal definition and complicate it. I'll update the ADR and mention that we will include the default config in all the producer applications as a starting point.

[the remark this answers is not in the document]
navinkarkera: Updated.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
robrap: I was actually thinking that the default configs would exist, but that they'd be disabled at first. I've really grown to not like commented out code (which I could explain more). This code feels more legitimate, even though it wouldn't actually do anything (for now) and you'd have to literally copy it to enable it. That said, in the future, by default we could simply enable events that the platform really is depending on.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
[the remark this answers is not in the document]
robrap: Sounds good to me.

sameenfatima78:
@mariajgrimaldi @mphilbrick211 @robrap Hi folks, would it be possible to get this merged soon? It will really help with the deployment of our project.

robrap:
@sameenfatima78: Navin will need to speak to the timeline. At the same time, temporarily adding the hardcoded version with a feature flag shouldn’t be a big deal and will keep you unblocked. Let me know if you want to discuss further.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,89 @@
...
+   #   Note: The topic names should not include environment prefix as it will be dynamically added based on
+   #   EVENT_BUS_TOPIC_PREFIX setting.
+   EVENT_BUS_PRODUCER_CONFIG = {
+       'org.openedx.content_authoring.xblock.deleted.v1': {
+           'topics': ['content-authoring-xblock-lifecycle'],
+           'event_key_field': 'xblock_info.usage_key',
+           'enabled': True,
+       },
+       'org.openedx.content_authoring.xblock.published.v1': {
+           'topics': ['content-authoring-xblock-lifecycle', 'xblock-published'],
```
robrap, line 45: Nit: could add content-authoring to the other topic name to better share the naming convention.

robrap (approved):

docs/decisions/0011-producing-to-event-bus-via-settings.rst
robrap: 1. If we go down the path of two settings, I think `EXTRA` is a term I've seen for `ADDITIONAL`.
2. We could also consider just overriding by keys like we do for some settings.

docs/decisions/index.rst
```diff
@@ -15,3 +15,4 @@ Architectural Decision Records (ADRs)
    0008-signals-with-pii
    0009-course-catalog-info-changed-design
    0010-multiple-event-types-per-topic
+   0011-producing-to-event-bus-via-settings
```
timmc-edx, line 18: I've just merged an ADR 11s, so you'll need to renumber before merge.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: > Then a deployer could just append to the end of the list, and there could be multiple entries for a single event, with different topics.

@timmc-edx Makes sense. Like the idea of having a list to easily add more topics.

> This still doesn't cover the case of "what if someone wants to exclude a topic for some event?" so it might not be exactly what we want. But I do think this is an issue that's likely to come up, and we should have a story for how are going to manage it.

How about using list of objects for each topic under an event, something like below:

```python
{
    'org.openedx.content_authoring.xblock.published.v1': [
        {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
        {'topic': 'xblock-published', 'event_key_field': 'xblock_info.usage_key', 'enabled': False},
    ],
    'org.openedx.content_authoring.xblock.deleted.v1': [
        {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
    ],
}
```
This would allow us to set different `event_key_field` for each topic as well as exclude some topics for some events without overriding the full base configuration.

robrap: @navinkarkera: I'm not clear on how the list can provide partial override without a `BASE` vs `EXTRA` setting. I do understand how we might take original per event type, unless an overridden event type is supplied. 

I'm fine with `event_key_field` being per topic, just in case we ever want something different.

navinkarkera: >@navinkarkera: I'm not clear on how the list can provide partial override without a BASE vs EXTRA setting. I do understand how we might take original per event type, unless an overridden event type is supplied.

Sorry I wasn't clear before. We will need `BASE` and `EXTRA`. I am proposing the format of the config for them. The `EXTRA` can override and disable or enable a event with same topic name from base config. One more option would be to have something like below which makes merging base and extra simple but the config doesn't look super nice.

```python
{
    'org.openedx.content_authoring.xblock.published.v1': {
        'content-authoring-xblock-lifecycle': {'event_key_field': 'xblock_info.usage_key', 'enabled': True},
        'xblock-published': {'event_key_field': 'xblock_info.usage_key', 'enabled': False},
    },
    'org.openedx.content_authoring.xblock.deleted.v1': {
        'content-authoring-xblock-lifecycle': {'event_key_field': 'xblock_info.usage_key', 'enabled': True},
    },
}
```

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Status
+******
+
+**Accepted** 2023-06-05
+
+Context
+*******
+
+The initial implementation of the event bus allowed only a single event type to be published to a topic, with details like topic name, consumer group name, and consumer name configured via code. We weren't sure what the api would finally look like, and whether the event bus config would live as part of the definition of the event, so we just went with explicit code that we could iterate upon.
```
mariajgrimaldi, line 12: can we please add a reference on what a topic is?

mariajgrimaldi, line 12:
```suggestion
The initial implementation of the event bus allowed only a single event type to be published to a topic, with details like topic name, consumer group name, and consumer name configured via code. We weren't sure what the API would finally look like, and whether the event bus config would live as part of the definition of the event, so we just went with explicit code that we could iterate upon.
```

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,89 @@
...
+.. code-block:: python
+
+   # .. setting_name: EVENT_BUS_PRODUCER_CONFIG
+   # .. setting_default: {}
+   # .. setting_description: Dictionary specifying event_types (keys) and its configuration dictionary as values.
+   #   Each configuration dictionary contains a flag denoting whether the event will be enabled or disabled,
+   #   topics/stream names where the event will be pushed to, event_key_field which is
+   #   and period delimited string path to event data field to use as event key.
+   #   Note: The topic names should not include environment prefix as it will be dynamically added based on
+   #   EVENT_BUS_TOPIC_PREFIX setting.
```
mariajgrimaldi, line 37: can we add a code inline annotation that specifies each field for the configurations? It looks a bit cluttered if it's part of the description.

mariajgrimaldi: There, we could specify the defaults if any

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,89 @@
...
+           connect_or_disconnect_handlers(configured_signal)
+
+The generic handler will again read the configuration and get details for event or signal triggered and push it to the event bus.
+
+.. code-block:: python
+
+   def general_signal_handler(sender, signal, **kwargs):
+       config = read_config(signal)
+       for topic in topics:
+           push_event_to_implementation(signal, config)
```
mariajgrimaldi, line 70: Can we specify what `push_event_to_implementation` would look like? A short description it's fine

docs/decisions/0011-producing-to-event-bus-via-settings.rst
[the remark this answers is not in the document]
mariajgrimaldi: Just a quick comment here: 
It might be fine to add additional arguments to the events' class definition -- but let's try not to make it that specific since Open edX Events is not necessarily tied to the event bus. Also, I don't see how _topics_ are related to the event definition itself like its data or its current version are. 

If we manage to find an argument justifying why other configurations should belong to the event definition. We could add a: `event_bus_defaults = {}` optional argument, which gives me enough context instead of just _topics_.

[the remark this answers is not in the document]
robrap: @mariajgrimaldi: We decided not to move this to the signal definition. See https://github.com/openedx/openedx-events/pull/224#discussion_r1220937831. This can be marked resolved.

mariajgrimaldi (approved):
I left a few tiny comments

docs/decisions/0011-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,89 @@
...
+
+.. code-block:: python
+
+   def ready(self):
+       load_all_signals()
+       config = read_config()
+       # validate the config dictionary and raise errors if any
+       validate_config()
+       for configured_signal in config:
+           connect_or_disconnect_handlers(configured_signal)
```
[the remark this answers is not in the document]
navinkarkera: If the plugin settings are added using [edx-django-utils](https://github.com/openedx/edx-django-utils/blob/master/edx_django_utils/plugins/plugin_settings.py), then yes. I tested it by adding openedx_events to installed apps in edx-platform lms settings and inside ready method of openedx-events, accessed settings from eox-tenant. It works.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: >  I'm even wondering if BASE vs EXTRA is more complicated than we want, and if we should have a simple setting, and expect it to be 100% overridden and explicit?

Agreed. We could go with the simple approach initially and upgrade if required. Just have a single setting dictionary as below:

```python
{
    'org.openedx.content_authoring.xblock.published.v1': [
        {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
        {'topic': 'xblock-published', 'event_key_field': 'xblock_info.usage_key', 'enabled': False},
    ],
    'org.openedx.content_authoring.xblock.deleted.v1': [
        {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
    ],
}
```

cc: @timmc-edx

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: Done.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: @timmc-edx @robrap What do you think about going ahead with the simple approach?

robrap: It seems like a reasonable place to start to me.

timmc-edx: Yes, I think starting with that makes sense.

People who want to override can do so carefully by copying the base config and then commenting out (rather than deleting) anything they want to omit, and adding a comment next to anything they add. It's not perfect, but it seems close enough.

It might become an issue once the event bus becomes a mandatory part of Open edX deployments (I think this will happen eventually) because then a mandatory signal/topic pair may be added to a field and deployers might not know to add it to their overrides. Maybe that's not a real problem, though! Our release notes would have to tell people to make the topics in advance, and that would also be an opportunity to tell them to ensure any overrides include the new topic.

navinkarkera: Thanks for the inputs. @timmc-edx @robrap 

Updated the ADR to reflect the changes in our discussion.

docs/decisions/0011-producing-to-event-bus-via-settings.rst
navinkarkera: @mariajgrimaldi The configuration part has changed a little. I am not sure on how to go about adding annotation inline inside the dictionary. Currently, it is like this:

```python
# .. setting_name: EVENT_BUS_PRODUCER_CONFIG
# .. setting_default: {}
# .. setting_description: Dictionary with event_types as keys and list of dictonaries for topic related configuration.
#    Each topic configuration dictionary contains a flag denoting whether the event will be published to the topic,
#    topic/stream name where the event will be pushed to, event_key_field which is and period delimited string path
#    to event data field to use as event key.
#    Note: The topic names should not include environment prefix as it will be dynamically added based on
#    EVENT_BUS_TOPIC_PREFIX setting.
EVENT_BUS_PRODUCER_CONFIG = {
    'org.openedx.content_authoring.xblock.published.v1': [
        {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
        {'topic': 'content-authoring-xblock-published', 'event_key_field': 'xblock_info.usage_key', 'enabled': False},
    ],
    'org.openedx.content_authoring.xblock.deleted.v1': [
        {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
    ],
}
```

By adding code inline annotation, do you mean some thing like:

```python
   # .. setting_name: EVENT_BUS_PRODUCER_CONFIG
   # .. setting_default: {}
   # .. setting_description: Dictionary with event_types as keys and list of dictonaries for topic related configuration.
   EVENT_BUS_PRODUCER_CONFIG = {
       'org.openedx.content_authoring.xblock.published.v1': [
        # .. setting_default: {}
        #    Each topic configuration dictionary contains a flag denoting whether the event will be published to the topic,
        #    topic/stream name where the event will be pushed to, event_key_field which is and period delimited string path
        #    to event data field to use as event key.
        #    Note: The topic names should not include environment prefix as it will be dynamically added based on
        #    EVENT_BUS_TOPIC_PREFIX setting.
           {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
           {'topic': 'xblock-published', 'event_key_field': 'xblock_info.usage_key', 'enabled': False},
       ],
       'org.openedx.content_authoring.xblock.deleted.v1': [
           {'topic': 'content-authoring-xblock-lifecycle', 'event_key_field': 'xblock_info.usage_key', 'enabled': True},
       ],
   }

```

robrap: When the annotations are extracted to an RST doc:
1. All the annotations must remain together to be parsed correctly.
2. The code will not be available in the doc, just the annotation text.
3. We can't invent new annotations without annotation definition changes that make them recognizable.

We can do something like the following though:
```
# .. setting_description: Dictionary with event_types as keys and a list of dictionaries for topic related configuration. 
#    Each dictionary in the list contains:
#      topic: <explain this>.
#      event_key_field: <explain this>.
#      ...
```
Note that the final RST has no formatting, so these indents would be lost if this were ever turned into a doc.

docs/decisions/0012-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,90 @@
...
+Decision
+********
+
+Create a generic signal handler to push events to the event bus. This handler should be attached to or connected to the signals that are enabled in Django settings. The configuration format will be as shown below:
+
+.. code-block:: python
+
+   # .. setting_name: EVENT_BUS_PRODUCER_CONFIG
+   # .. setting_default: {}
+   # .. setting_description: Dictionary with event_types as keys and list of dictonaries for topic related configuration.
```
timmc-edx, line 34: [nit] Typo, and suggested rewording.

```suggestion
   # .. setting_description: Dictionary of event_types to lists of dictionaries for topic related configuration.
```

docs/decisions/0012-producing-to-event-bus-via-settings.rst
```diff
@@ -0,0 +1,90 @@
...
+   #    Each topic configuration dictionary contains a flag denoting whether the event will be published to the topic,
+   #    topic/stream name where the event will be pushed to, event_key_field which is and period delimited string path
```
timmc-edx, line 36: [nit] typo, hyphenation
```suggestion
   #    topic/stream name where the event will be pushed to, event_key_field which is a period-delimited string path
```

docs/decisions/0011-producing-to-event-bus-via-settings.rst
timmc-edx: I'm pretty happy with what we have now -- I think it just needs explicit naming of the fields (`enabled`, `topic`) and it's ready to go, as far as I'm concerned.

navinkarkera: Updated with explicit names in description.

navinkarkera:
This is ready to be merged now.

timmc-edx (approved):

openedx-webhooks:
@navinkarkera 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-events&entry.867055334=https://github.com/openedx/openedx-events/pull/224&entry.1484655318=https://github.com/navinkarkera&entry.752974735=2023-05-23+06:29&entry.1917517419=2023-06-26+14:12&entry.2133058324=Yes) so we can improve your experience in the future.

PR #225, opened 2023-05-24
feat: add unit tests around schema evolution

rgraber:
**Description:**
Add unit tests to confirm that any schema updates are backward and forward compatible with stored versions. Also adds a management command to generate and store the current Avro schema for each signal.

Manual testing:
Test 1: Add required field "thing" to CourseEnrollmentData
`FAILED openedx_events/event_bus/avro/tests/test_avro.py::TestAvro::test_evolution_is_backward_compatible - fastavro._read_common.SchemaResolutionError: No default value for thing`

Test 2: Remove required field "creation_date" from CourseEnrollmentData
`FAILED openedx_events/event_bus/avro/tests/test_avro.py::TestAvro::test_evolution_is_forward_compatible - fastavro._read_common.SchemaResolutionError: No default value for creation_date`

Test 3: Change type of user_data from UserData to str in CourseEnrollmentData
`FAILED openedx_events/event_bus/avro/tests/test_avro.py::TestAvro::test_evolution_is_backward_compatible - fastavro._read_common.SchemaResolutionError: Schema mismatch: {'type': 'record', 'name': 'org.openedx.learning.course.enrollment.created.v1.UserData', 'fields': [{'name': 'id', 'ty...`
`FAILED openedx_events/event_bus/avro/tests/test_avro.py::TestAvro::test_evolution_is_forward_compatible - fastavro._read_common.SchemaResolutionError: Schema mismatch: string is not {'type': 'record', 'name': 'org.openedx.learning.course.enrollment.created.v1.UserData', 'fields': [{'na...`

Test 4: Add optional field thing to CourseEnrollmentData
All tests pass

Test 4: Remove optional field created_by from CourseEnrollmentData
All tests pass

**ISSUE:** https://github.com/edx/edx-arch-experiments/issues/271
**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/openedx_events/event_bus/avro/tests/test_avro.py
+++ b/openedx_events/event_bus/avro/tests/test_avro.py
@@ -197,10 +197,7 @@ class TestAvro(FreezeSignalCacheMixin, TestCase):
 
         Caveat: While we can go from an attrs class to a schema, we cannot go from a schema to an attrs class in the
         same way because of custom serializers. Thus, when we are generating test data, we can only generate
-        it using the latest version of the attrs class, and not one derived from the stored schema. This means this
-        test does not perfectly reflect what would happen if the consumer was updated before the producer, but it still
-        fails if the latest attrs class is not backward compatible with the old schema. It will just fail on writing
-        instead of reading.
+        the dictionary we would get after calling schema.to_dict.
         """
         for signal in OpenEdxPublicSignal.all_events():
             if signal.event_type in KNOWN_UNSERIALIZABLE_SIGNALS:
--- a/openedx_events/event_bus/avro/tests/test_avro.py
+++ b/openedx_events/event_bus/avro/tests/test_avro.py
@@ -1,8 +1,12 @@
 """Test interplay of the various Avro helper classes"""
+import io
+import os
 from datetime import datetime
 from typing import List
 from unittest import TestCase
 
+from fastavro import schemaless_reader, schemaless_writer
+from fastavro.schema import load_schema
 from opaque_keys.edx.keys import CourseKey, UsageKey
 
 from openedx_events.event_bus.avro.deserializer import AvroSignalDeserializer, deserialize_bytes_to_event_data
@@ -16,15 +20,7 @@ from openedx_events.event_bus.avro.tests.test_utilities import (
     create_simple_signal,
 )
 from openedx_events.tests.utils import FreezeSignalCacheMixin
-from openedx_events.tooling import OpenEdxPublicSignal, load_all_signals
-
-# If a signal is explicitly not for use with the event bus, add it to this list
-#  and document why in the event's annotations
-KNOWN_UNSERIALIZABLE_SIGNALS = [
-    "org.openedx.learning.discussions.configuration.changed.v1",
-    "org.openedx.content_authoring.course.certificate_config.changed.v1",
-    "org.openedx.content_authoring.course.certificate_config.deleted.v1",
-]
+from openedx_events.tooling import KNOWN_UNSERIALIZABLE_SIGNALS, OpenEdxPublicSignal, load_all_signals
 
 
 def generate_test_event_data_for_data_type(data_type):
@@ -40,7 +36,6 @@ def generate_test_event_data_for_data_type(data_type):
     Returns:
         (dict): A data dictionary containing dummy data for all attributes of the class
     """
-    data_dict = {}
     defaults_per_type = {
         int: 1,
         bool: True,
@@ -53,6 +48,7 @@ def generate_test_event_data_for_data_type(data_type):
         List[int]: [1, 2, 3],
         datetime: datetime.now(),
     }
+    data_dict = {}
     for attribute in data_type.__attrs_attrs__:
         result = defaults_per_type.get(attribute.type, None)
         if result is not None:
@@ -65,8 +61,29 @@ def generate_test_event_data_for_data_type(data_type):
     return data_dict
 
 
+def generate_test_data_for_signal(signal: OpenEdxPublicSignal) -> dict:
+    """
+    Generates test data for use in the event bus test cases.
+
+    Builds data by filling in dummy data for basic data types (int/float/bool/str)
+    and recursively breaks down the classes for nested classes into basic data types.
+
+    Arguments:
+        data_type: The type of the data which we are generating data for
+
+    Returns:
+        (dict): A data dictionary containing dummy data for all attributes of the class
+    """
+    test_data = {}
+    for key, curr_class in signal.init_data.items():
+        example_data = generate_test_event_data_for_data_type(curr_class)
+        example_data_processed = curr_class(**example_data)
+        test_data.update({key: example_data_processed})
+    return test_data
+
+
 class TestAvro(FreezeSignalCacheMixin, TestCase):
-    """Tests for end-to-end serialization and deserialization of events"""
+    """Tests for end-to-end serialization and deserialization of events and schema evolution"""
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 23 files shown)


robrap (commented):
* I wonder if we need a doc that explains the process for updating a schema, even if it is just what we know today? Could be docstring, as long as someone would know where to look, so we might just have a README.rst that points to the right docstring.
* Can you do a couple of manual tests and list what you tested? E.g. making an incompatible change to the signal before and after updating the schemas. Making a compatible change, etc. 
* Also, I wonder if the management command should do some of this testing and complain if you are trying to make an incompatible change, unless you provide some "force" argument, in case you are first working on the event. Or I guess you could just delete the schema. 

Thoughts to be discussed.

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -78,17 +95,80 @@ def test_all_events(self):
...
+
+    def test_evolution_is_backward_compatible(self):
+        """
+        Test current version of events is backward compatible with stored schemas.
+
+        There's no assert because as long as the test doesn't raise an error, the new schema is
+        backward-compatible with the original.
+
+        Caveat: While we can go from an attrs class to a schema, we cannot go from a schema to an attrs class in the
+        same way because of custom serializers. Thus, when we are generating test data, we can only generate
```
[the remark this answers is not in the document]
rgraber: It's possible we could change how we serialize custom objects to give us a trail but I don't think it's worth it for this one test.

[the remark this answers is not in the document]
robrap: Agreed. Forget about this. Thanks for explaining.

rgraber:
> * I wonder if we need a doc that explains the process for updating a schema, even if it is just what we know today? Could be docstring, as long as someone would know where to look, so we might just have a README.rst that points to the right docstring.
> 

I would put that in a separate documentation-only PR. Among other things, there will probably be some debate on where to put it.

>     * Can you do a couple of manual tests and list what you tested? E.g. making an incompatible change to the signal before and after updating the schemas. Making a compatible change, etc.

Done
> 
>     * Also, I wonder if the management command should do some of this testing and complain if you are trying to make an incompatible change, unless you provide some "force" argument, in case you are first working on the event. Or I guess you could just delete the schema.

I thought we didn't really want people overwriting existing schemas regardless of whether or not the change is compatible, so we're always testing against the original version of the schema. 

> 
> 
> Thoughts to be discussed.

openedx_events/management/commands/generate_avro_schemas.py
```diff
@@ -0,0 +1,89 @@
...
+class Command(BaseCommand):
+    """
+    Management command to save current Avro schemas for all signals for schema evolution testing.
+    """
+
+    help = """
+    Generate and save the current Avro schemas of OpenEdxPublicSignals.
+
+    Schemas will be saved in openedx_events/avro/tests/schemas. The folder will be created if it does not already exist.
+    Should only be used for new signals. Existing files will be overwritten, so use with caution.
```
robrap, line 26: Thoughts on making it easier to do the right thing:
```suggestion
    Should only be used for new signals. If a schema is changing, it is recommended you leave the original
    in place. Only overwrite the file for an exceptional case, like the schema still being under development.
```

openedx_events/management/commands/generate_avro_schemas.py
```diff
@@ -0,0 +1,89 @@
...
+                logger.info(f"Known unserializable signal: {signal.event_type}. Skipping.")
+                continue
+            serializer = AvroSignalSerializer(signal)
+            schema_dict = serializer.schema
+            filename = f"{signal.event_type.replace('.','+')}_schema.avsc"
+            root_path = import_module('openedx_events').__path__[0]
+            folder_path = f"{root_path}/event_bus/avro/tests/schemas"
+            full_file_name = f"{folder_path}/{filename}"
+            if os.path.exists(full_file_name):
+                confirmation = input(f"Warning: overwriting schema for {signal.event_type}. Are you sure you want to "
```
robrap, line 80: * At a minimum, can we make our recommendation more clear? For example:
```suggestion
                confirmation = input(f"Warning: Overwriting schema for {signal.event_type}. We recommend choosing 'n' to leave the original schema in place? Do you want to continue anyway [y/n]?"
```
* An alternative is to warn and skip, and add docstring + message that the original schema should be deleted if you want the management command to recreate it.

robrap (commented):
This is really great. I'm excited about it. Note that there is a group considering consuming messages using node, and this Avro schemas will come in handy. :)

robrap:
> I thought we didn't really want people overwriting existing schemas regardless of whether or not the change is compatible, so we're always testing against the original version of the schema. That's also why we always have an "are you sure" prompt when overwriting.

Sounds reasonable. I made other PR comments to see if we could add this context to your code comments, but otherwise consider this thread resolved. Thanks.

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -16,18 +20,66 @@
...
 
-# If a signal is explicitly not for use with the event bus, add it to this list
-#  and document why in the event's annotations
-KNOWN_UNSERIALIZABLE_SIGNALS = [
-    "org.openedx.learning.discussions.configuration.changed.v1",
-    "org.openedx.content_authoring.course.certificate_config.changed.v1",
-    "org.openedx.content_authoring.course.certificate_config.deleted.v1",
-]
+
+def generate_test_data_for_schema(schema):  # pragma: no cover
```
robrap, line 26: Curious why `# pragma: no cover` is needed for this and the other method, when it seems like they are used in tests?

robrap (commented):
Minor questions.

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -78,17 +95,80 @@ def test_all_events(self):
...
+
+            # write to bytes using current schema
+            current_out = io.BytesIO()
+            data_dict = serializer.to_dict(test_data)
+            schemaless_writer(current_out, schema_dict, data_dict)
+            current_out.seek(0)
+            current_event_bytes = current_out.read()
+
+            # get stored schema
+            stored_schema = load_schema(f"{os.path.dirname(os.path.abspath(__file__))}/schemas/"
```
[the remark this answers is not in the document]
rgraber: We don't have a full process for updating schemas so I'm hesitant to make much documentation about it. In any case, if there's no file, this would be an issue with how schemas are added as opposed to updated. I guess we can document that somewhere.

openedx_events/event_bus/avro/tests/test_avro.py
robrap: I see. You could add the pragma more specifically on the code that isn't actually covered, but this is non-blocking. It just raised questions for me. Thanks.

openedx_events/event_bus/avro/tests/test_avro.py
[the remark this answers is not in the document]
robrap: Could you even just add a comment above this line that if the schema is missing, they probably need to run the management command to generate a file for a new event?

openedx_events/event_bus/avro/tests/test_avro.py
rgraber: I thought it worthwhile to future-proof. Coverage shouldn't really care about this method at all, and if someone changes it they shouldn't have to go through the annoyance of figuring out which bits are and aren't covered again.

openedx_events/event_bus/avro/tests/test_avro.py
```diff
@@ -78,17 +152,86 @@ def test_all_events(self):
...
+            serializer = AvroSignalSerializer(signal)
+            schema_dict = serializer.schema
+
+            # get stored schema
+            schema_filename = f"{os.path.dirname(os.path.abspath(__file__))}/schemas/" \
+                              f"{signal.event_type.replace('.','+')}_schema.avsc"
+            try:
+                old_schema = load_schema(schema_filename)
+            except SchemaRepositoryError:  # pragma: no cover
+                self.fail(f"Missing file {schema_filename}. If a new signal has been added, you may need to run the"
```
robrap, line 220: Thanks for this. We know this is an expected situation, and it just makes the experience much nicer for adding new events.

Issue #209, opened 2023-05-01
Make ADR for versioning/dependency strategy around event bus implementations


timmc-edx:
A/C

- [x] An ADR explaining why we're going to depend on all implementations of the event bus: https://github.com/openedx/openedx-events/pull/227
- [x] An issue for doing this better in the future (possibly an Axim issue)
  - Declined -- we're aware of the broader plugin dependencies issue and anything we do to solve that will help with the event bus implementations.

We reached a working consensus that we would depend on all implementations from any IDA that needed to use the Event Bus, but never made an ADR out of it: https://github.com/openedx/openedx-events/issues/88#issuecomment-1322556980

Let's finish that process.

robrap:
@timmc-edx: Reminder to hopefully close this out. Thank you!

Issue #88, opened 2022-08-04
Determine a dependency versioning strategy for event-bus implementations


timmc-edx:
The `edx_event_bus_kafka` dependency will be specified and pinned in a file _outside_ of each IDA that uses it, since openedx-events will load the implementation by looking up a module reference in the IDA's config. This means that dependency version changes on the library will not be versioned with the IDA (i.e. in the requirements/base.txt files) and we'll need some sort of expand-contract approach to API changes.

For example, we might need to do something like this:

1. Add a new `KafkaProducer` class in event-bus-kafka and delegate the existing `send_to_event_bus` to use it
2. Switch the producer code in Studio to use the new class
3. Delete the `send_to_event_bus` function from event-bus-kafka

Every master-branch deployer would need to do this in lockstep. (Maybe that's just 2U?) For Open edX releases we might be able to get away with just saying "use the release tag".

## Acceptance Criteria

- [ ] Design a strategy that will work for deployers in general
- [ ] Document how-to for these upgrades in appropriate places.

timmc-edx:
It turns out we have a collection of related dependency-related issues:
- A bunch of our repos will depend on event-bus-kafka but won't be able to declare it directly, and event-bus-kafka has its own dependencies that might cause deploy-time version conflicts.
    - As Robert [notes](https://github.com/edx/edx-arch-experiments/issues/52#issuecomment-1204522831), we won't be able to use `make upgrade` to keep things updated or synchronized (unless we change how deployments work).
- We'll have to do expand-contract deploys with event-bus-kafka as a result, and won't have safeguards. Maybe the same for all other deployers, eventually.
- There's a weird sort-of-bidirectional dependency between openedx-events and event-bus-kafka: The second declares a dependency on the first, but the first has an interface dependency on the second.

One thought I had is that we might just want to pack event-bus-kafka _into_ openedx-events as a PEP-508 "extras" package, so we can include `openedx-events[kafka]` in base.in. (And eventually have an extra for each implementation.) Or maybe rewrite the dependency files to add the `[kafka]` part at AMI build time? It doesn't solve all the problems, but it would at least allow the interface to be versioned.

(You could even have a special CI workflow file on IDAs that confirms that none of the various extras' dependencies would conflict with the IDA's compiled requirements files.)

robrap:
[...]

PR #227, opened 2023-05-25
docs: Add ADR for depending on multiple event bus implementations

timmc-edx:
Issue: https://github.com/openedx/openedx-events/issues/209

- [x] Re-read comments of https://github.com/openedx/openedx-events/issues/88 to see if there are things I've missed
- [x] Before merge, check if Navin's ADR has landed yet (this might need to be renumbered from 11 to 12)

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
+++ b/docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
@@ -15,6 +15,8 @@ The Event Bus is an abstraction allowing events to be passed from one IDA to ano
 
 Any given deployer will likely only be using one of these implementations at most, and will need to install it one way or another. This can either be by explicit inclusion in the base requirements of each IDA ("in-tree") or by a separate installation process of one sort or another ("out-of-tree"). Each implementation package also has its own dependencies and dependency version constraints which may conflict with the other dependencies pulled in by any given IDA. Each package will need to be kept compatible with a significant number of IDAs.
 
+Currently, use of the event bus is optional, so choice of implementation is moot for most deployers. However, we expect that the event bus will later become a core part of the architecture, such that all deployments will need to use one implementation or another.
+
 Decision
 ********
 
@@ -40,8 +42,8 @@ There are also some downsides:
 
 We are not necessarily *satisfied* with this approach, but regard it as suitable for at least one release.
 
-Rejected Alternatives
-*********************
+Rejected or deferred alternatives
+*********************************
 
 There are several alternative approaches we have considered:
 
@@ -67,12 +69,16 @@ As discussed, this results in more development friction (although possibly lower
 
 Not having *any* implementation provided in a standard devstack installation also makes Event Bus work unnecessarily difficult.
 
+This decision is taking place against a backdrop of trying to use plugins and other extension mechanisms in more places, which would frequently require external dependencies. In the event bus work we have deferred figuring out how to reconcile out-of-tree dependency lists against our upgrade and pinning processes, but we'll need to figure it out sooner or later. Once we do, this option becomes more palatable (although likely not as appealing as the option in the next section, "one default").
+
 One default
 ===========
 
 Another option would be to select one implementation to always keep in the dependencies, and require that others be installed out-of-tree. If we went this route, we would likely choose event-bus-redis as the default, as it does not require the deployer to run additional services. (Redis is already used in Open edX deployments.)
 
 Given that event-bus-kafka already requires out-of-tree installation, this may be a path we take in the future. However, 2U is using event-bus-kafka and developed the initial version of the event bus, and the direct inclusion of the dependency made that work easier. As the event bus and its main implementations stabilize, this factor may no longer have much weight.
+
+As of May 2023, we are moving towards including event-bus-redis in *development* settings so that the event bus can be used in a standard devstack. It is possible that we will later choose to extend this to other environments, such that Redis will be the default for deployment as well. If so, we will likely reconsider our decision to include the Kafka implementation in our dependencies by default.
 
 References
 **********
--- a/docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
+++ b/docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
@@ -4,9 +4,9 @@
 Status
 ******
 
-**Accepted** *2023-05-25*
+**Accepted** *2023-06-09*
 
-(This decision was actually made in November 2022, and retroactively documented in May 2023 but with additional, newer context such as the existence of event-bus-redis.)
+(This decision was actually made in November 2022 during the development of event-bus-kafka, and retroactively documented in June 2023 but with additional, newer context such as the existence of event-bus-redis.)
 
 Context
 *******
@@ -16,6 +16,10 @@ The Event Bus is an abstraction allowing events to be passed from one IDA to ano
 Any given deployer will likely only be using one of these implementations at most, and will need to install it one way or another. This can either be by explicit inclusion in the base requirements of each IDA ("in-tree") or by a separate installation process of one sort or another ("out-of-tree"). Each implementation package also has its own dependencies and dependency version constraints which may conflict with the other dependencies pulled in by any given IDA. Each package will need to be kept compatible with a significant number of IDAs.
 
 Currently, use of the event bus is optional, so choice of implementation is moot for most deployers. However, we expect that the event bus will later become a core part of the architecture, such that all deployments will need to use one implementation or another.
+
+`OEP-52`_ contains links to additional context and ADRs, including the decisions to use Kafka and Redis.
+
+.. _OEP-52: https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0052-arch-event-bus-architecture.html
 
 Decision
 ********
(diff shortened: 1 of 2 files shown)


docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
```diff
@@ -0,0 +1,72 @@
...
+
+There are also some downsides:
+
+- The dependency tree is larger, and is more or less guaranteed to include at least one top-level dependency that each deployer is not using. Not only does each dependency take up bandwidth to download, space on disk, and time spent in upgrading packages, but it can also bring licensing and architecture incompatibilities.
+- More concretely, event-bus-kafka depends on confluent-kafka, but that package does not provide binaries compatible with Linux on Apple Silicon. This means that developers on Apple M1 laptops cannot install this package in devstack at all. So as a knock-on effect, event-bus-kafka has had to keep confluent-kafka out of its own in-tree dependencies; deployers wishing to use it have to install and version that package out-of-tree anyhow. This obviates a number of the stated benefits for development and version management, at least for the Kafka users.
```
bradenmacdonald, line 37: Presumably this will be fixed at some point (there are PRs open for it already upstream), and then the benefits of this approach will be restored, right?

timmc-edx: True! I'll make a note that it might be a temporary situation. (I'm not holding my breath, though -- the librdkafka folks don't seem interested in finding an alternative CI provider that supports Apple Silicon.)

docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Implement inside openedx-events
+===============================
+
+We briefly discussed the idea of including all of the implementations *in* the openedx-events package, possibly as extras (e.g. ``openedx-events[kafka]``). However, this would have complicated the development of openedx-events; it would have become impossible to e.g. pin the version of the Redis implementation in a dependent IDA while continuing to receive updates to the Kafka implementation. (The use of extras also would not have helped our dependencies issue; we'd then need to decide which extras to depend upon from IDAs' base requirements files.)
+
+Single implementation
+=====================
+
+It's very unlikely that one implementation would satisfy everyone. Any event store, queue, or message bus technology that is used is additional software that deployers must learn to install, manage, configure, and operate. The exceptions would be MySQL and Redis, which are already used in Open edX. However, neither of these was used for the first implementation. Therefore, even if Redis turns out to work very well, we'll still need the extension point.
```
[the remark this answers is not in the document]
bradenmacdonald: According to [the ADR](https://github.com/openedx/event-bus-kafka/blob/main/docs/decisions/0002-kafka-based-event-bus.rst), the cons for Redis were that it can "Can lose acked data, even if RAM backed up with an append-only file (AOF)." (unfortunately no details were provided, so I can't tell if this claim is still relevant), and that it "Requires homegrown schema management." (which I think we are doing anyways?)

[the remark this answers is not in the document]
timmc-edx: Hah! Of course I should have checked previous ADRs, thank you. :-)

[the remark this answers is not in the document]
robrap: A more current ADR also exists for why redis (vs why not redis): https://github.com/openedx/event-bus-redis/blob/main/docs/decisions/0002-redis-streams-based-event-bus.rst

docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
```diff
@@ -0,0 +1,80 @@
...
+===============
+
+As discussed, this results in more development friction (although possibly lower maintenance burden). The package also would not participate in our usual ``make upgrade`` process. This leads to either A) more operational risk if deployers leave the implementation package version unpinned in their deployment pipeline, or B) more toil if the version is pinned but has to be manually bumped in every IDA every time a new package version is released.
+
+Not having *any* implementation provided in a standard devstack installation also makes Event Bus work unnecessarily difficult.
+
+One default
+===========
+
+Another option would be to select one implementation to always keep in the dependencies, and require that others be installed out-of-tree. If we went this route, we would likely choose event-bus-redis as the default, as it does not require the deployer to run additional services. (Redis is already used in Open edX deployments.)
```
robrap, line 73: I wonder if it makes sense to make a redis-as-default for development as a decision as part of this ADR? It wouldn't necessarily change what we are doing, but it would change how we write about it and how permanent this part of the decision might be. Even if we use a single ADR, maybe we make it more clear that deciding for kafka and redis are separate decisions, and we might be choosing to include them for different reasons (although potentially overlapping reasons as well).

docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Single implementation
+=====================
+
+It's very unlikely that one implementation would satisfy everyone. Any event store, queue, or message bus technology that is used is additional software that deployers must learn to install, manage, configure, and operate. The exceptions would be MySQL and Redis, which are already used in Open edX. However, neither of these was used for the first implementation. Therefore, even if Redis turns out to work very well, we'll still need the extension point.
+
+All out-of-tree
+===============
+
+As discussed, this results in more development friction (although possibly lower maintenance burden). The package also would not participate in our usual ``make upgrade`` process. This leads to either A) more operational risk if deployers leave the implementation package version unpinned in their deployment pipeline, or B) more toil if the version is pinned but has to be manually bumped in every IDA every time a new package version is released.
```
robrap, line 66: I wonder if this is temporarily-rejected/deferred? Could we find other solutions that improve all of these issues **and** keep us from installing unused packages for people? I think it would take a bunch of work, but as we want to better support plugins and optional parts of the platform, improved tools on this front might be worth the investment.

What I am saying does not change the decision, but may just change how we discuss this rejection. My notes shouldn't replace anything you wrote, but might add to it.

docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
```diff
@@ -0,0 +1,80 @@
...
+
+There are also some downsides:
+
+- The dependency tree is larger, and is more or less guaranteed to include at least one top-level dependency that each deployer is not using. Not only does each dependency take up bandwidth to download, space on disk, and time spent in upgrading packages, but it can also bring licensing and architecture incompatibilities.
+- More concretely, event-bus-kafka depends on confluent-kafka, but that package does not provide binaries compatible with Linux on Apple Silicon. This means that developers on Apple M1 laptops cannot install this package in devstack at all. So as a knock-on effect, event-bus-kafka has had to keep confluent-kafka out of its own in-tree dependencies; deployers wishing to use it have to install and version that package out-of-tree anyhow. While this may eventually be resolved, this does obviate a number of the stated benefits for development and version management for the Kafka users.
```
robrap, line 39: I thought that was something about this as "context" that made it especially painful when we were trying to install the packages optionally? Maybe I am misremembering? If you can recall, maybe calling that out as part of the context for making the decision would be useful. If not, please ignore.

timmc-edx: I think that we discovered the Apple Silicon problem after we had already started installing edx-event-bus-kafka in Ansible, so from the perspective of a somewhat retroactive ADR it's more Consequence than Context.

docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
timmc-edx: Yes, it's probably worth calling out that this is something we'd like to be able to do better.

docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
timmc-edx: Added a bit about default-for-development.

docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
```diff
@@ -0,0 +1,86 @@
+11. Depending on multiple event bus implementations
+###################################################
+
+Status
+******
+
+**Accepted** *2023-05-25*
+
+(This decision was actually made in November 2022, and retroactively documented in May 2023 but with additional, newer context such as the existence of event-bus-redis.)
```
robrap, line 9: Consider the following:
```suggestion
(This decision was actually made in November 2022 with event-bus-kafka, and retroactively documented in May 2023 but with additional, newer context such as the existence of event-bus-redis.)
```

docs/decisions/0011-depending-on-multiple-event-bus-implementations.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Context
+*******
+
```
robrap, line 13: Consider adding a link to the OEP for additional context of Kafka and Redis, since the ADRs are available from the OEP. https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0052-arch-event-bus-architecture.html

robrap (approved):
Added comments to consider, but non-blocking.

PR #267, opened 2023-09-19
feat: add/remove course staff role events

ilee2u:
**Description:**  

We are developing a new backend for exams called `edx-exams`. This backend will be utilized by instructor dashboards to interact with student exam attempts.

Circa Sept 2023, `edx-exams` has no way of knowing if a user is course staff (i.e. an instructor) or not. We'd like to avoid having this IDA call `edx-platform` to get this info (which would create a circular dependency), so we've decided to emit this information from `edx-platform` to `edx-exams` via the event bus (logic for the respective event producers and consumers to be deliberated later).

**This PR adds event definitions for adding and removing the course staff. These events are intended to be consumed by edx-exams, but they can also be consumed in other Open edX IDAs in the future as needed.**

**ISSUE:** [MST-1918](https://2u-internal.atlassian.net/browse/MST-1918) (Internal)

In order to receive course staff info in edx-exams, we must first define events to be sent. The events should be added to https://github.com/openedx/openedx-events/blob/main/openedx_events/learning/signals.py - there should be one event for adding a course staff role, and one event for removing (or revoking) a course staff role. The data for both events should consist of a user id and a course id.

**Dependencies:** None

**Merge deadline:** None (for now)

**Testing instructions:**

1. Test as usual

**Reviewers:**
- [x] @Zacharis278

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.


openedx_events/learning/signals.py
```diff
@@ -196,3 +197,25 @@
         "notification_data": UserNotificationData,
     }
 )
+
+# .. event_type: org.openedx.learning.user.add.course.staff.role.v1
+# .. event_name: ADD_COURSE_STAFF_ROLE
+# .. event_description: Assigns the "Course Staff" role to a user.
```
Zacharis278, line 203: I think this description is off because this event doesn't cause a course staff assignment, it's fired as a result of one. Might also want to tweak the event_name to be more clear on that as well.

Zacharis278: can you update the descriptions too?

Zacharis278:
my recent comment got collapsed as resolved but can you update the descriptions to indicate this is fired as a result of assigning course staff not a signal to assign course staff.

Zacharis278 (approved):
looks good 👍

openedx_events/learning/data.py
```diff
@@ -296,3 +296,17 @@ class ProgramCertificateData:
     status = attr.ib(type=str)
     url = attr.ib(type=str)
     certificate_available_date = attr.ib(type=datetime, default=None)
+
+
+@attr.s(frozen=True)
+class CourseStaffData:
+    """
+    Attributes defined for the Open edX Authentication function to Add Users to a role.
```
mariajgrimaldi, line 304: can we be less specific here? It could say something like: Attributes used to represent course staff. or something like it but less specific on how it'd be used.

openedx_events/learning/data.py
```diff
@@ -296,3 +296,17 @@ class ProgramCertificateData:
...
+
+    Arguments:
+        course_key (Course ID): identifier of the course where the user will have staff role assigned/removed
+        user (UserData): User who will have the "Course Staff" role assigned/removed
+    """
+
+    course_key = attr.ib(type=str)
+    user = attr.ib(type=UserData)
```
mariajgrimaldi, line 312: which other fields do you think would be useful?

mariajgrimaldi (commented):
Just a few minor comments.

openedx_events/learning/data.py
ilee2u: I added some other fields based on the current parameters that CourseAccessRole currently takes: https://github.com/openedx/edx-platform/blob/master/common/djangoapps/student/roles.py#L448

Please note that this is subject to change in the future due to the concurrent roles/permissions project that is being done (see the "IMPORTANT:" comment in the event data for more details).

ilee2u:
Quickly reiterating what I've written in the event data docstring about this being tech debt b/c of planned changes to roles and permissions:

edX currently uses roles, and only roles, to decide what kind of access a user has. There is an ongoing project to replace this roles-only system with a system that uses roles that are made up of permissions, which is being worked on in parallel with another project to emit events whenever users are assigned any type of "Course Staff" role.

It's unclear what the state of this roles/permissions project will be the time the events project is completed, so each project's respective teams will stay in touch with each other.

For now, we're making a best effort to publish this an event that will regard the permission(s) we'd expect to "filter" for in the future (For more info, please check out this document: https://docs.google.com/spreadsheets/d/1htsV0eWq5-y96DZ5A245ukfZ4_qeH0KjHVaOyfqD8OA/edit#gid=908503896) and not for the roles we have now. Likely this/these permission(s) will be something like `manage_students`, but we need to evaluate how this will align with some possible future roles such as `limited_staff` or `ccx`.

As such, the current plan is to do one of the following once the roles/permissions project's feature branch is merged to master:
    1. Modify this event to "filter" by the correct permissions once the
    2. As a backup plan, make a new event if this proves too difficult.

Until either of these plans are executed, the comment under the IMPORTANT header should stay put.

openedx_events/learning/data.py
ilee2u: I've changed the name to `ManageStudentsPermissionsData` to reflect planned changes to roles/permissions.

openedx_events/learning/data.py
mariajgrimaldi: That's better. Thank you!

mariajgrimaldi (approved):
These are really useful events. Thank you!

mariajgrimaldi:
I'll wait for the other PR to be merged.

mariajgrimaldi:
@ilee2u: the other PR was already merged!

docs/decisions/0012-producing-to-event-bus-via-settings.rst
```diff
@@ -1,4 +1,4 @@
-11. Enable producing to event bus via settings
+12. Enable producing to event bus via settings
```
mariajgrimaldi, line 1: Thank you for this! I totally missed it.

mariajgrimaldi:
Changes release. Thank you!

PR #268, opened 2023-09-19
Events for exams IDA

ilee2u:
**Description:** We are developing a new backend for exams called edx-exams. This backend will be utilized by instructor dashboards to interact with student exam attempts, and will store data regarding those attempts.

We are currently working to implement the downstream effects triggered whenever an exam attempt is submitted or reviewed. For example, when an exam attempt is submitted, we will want to make sure edx-platform knows to mark the exam subsection.

We have added several event definitions based on different actions that can be performed on exam attempts:

- EXAM_ATTEMPT_SUBMITTED
- EXAM_ATTEMPT_REJECTED
- EXAM_ATTEMPT_VERIFIED
- EXAM_ATTEMPT_ERRORED
- EXAM_ATTEMPT_RESET

These events will be consumed by various services in edx-platform in order to perform tasks such as creating grade overrides, invalidating certificates, changing credit requirement statuses, and more.

**ISSUE:** [MST-2083](https://2u-internal.atlassian.net/browse/MST-2083) (Internal)

**Dependencies:** None.

**Installation instructions:** List any non-trivial installation
instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen
4. If D happened instead - check failed.

**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.


openedx_events/learning/data.py
```diff
@@ -296,3 +296,23 @@ class ProgramCertificateData:
...
+
+    Arguments:
+        user_id (int): identifier of the user to which the grade override belongs
+        course_key (CourseKey): identifier of the course to which the grade override belongs
+        usage_key (UsageKey): identifier of the content that will get a grade override
+        requesting_user (UserData): user triggering event, sent only when instructor modifies an exam attempt
+        credit_requirement_status (str): status to change the student's CreditRequirement to
+    """
+
+    user_id = attr.ib(type=int)
```
Zacharis278, line 314: I noticed @MichaelRoytman's PR uses the UserData object here. Do you think we should use that?

openedx_events/learning/signals.py
```diff
@@ -196,3 +197,58 @@
...
+EXAM_ATTEMPT_SUBMITTED = OpenEdxPublicSignal(
+    event_type="org.openedx.learning.exam.attempt.submitted.v1",
+    data={
+        "exam_attempt_data": ExamAttemptData,
+    }
+)
+
+# .. event_type: org.openedx.learning.exam.attempt.rejected.v1
+# .. event_name: EXAM_ATTEMPT_REJECTED
+# .. event_description: Emitted when an exam attempt is marked rejected by an instructor or staff.
```
Zacharis278, line 214: I'd leave off "by an instructor or staff" in case this no longer becomes true in the future. (AI review by an tool as an example)

openedx_events/learning/signals.py
```diff
@@ -196,3 +197,58 @@
...
+EXAM_ATTEMPT_REJECTED = OpenEdxPublicSignal(
+    event_type="org.openedx.learning.exam.attempt.rejected.v1",
+    data={
+        "exam_attempt_data": ExamAttemptData,
+    }
+)
+
+# .. event_type: org.openedx.learning.exam.attempt.verified.v1
+# .. event_name: EXAM_ATTEMPT_VERIFIED
+# .. event_description: Emitted when an exam attempt is marked verified by an instructor or staff.
```
Zacharis278, line 225: same as the above comment. Right now an LTI tool can automatically trigger this if there is no suspicious behavior.

Zacharis278 (commented):
One comment on how we send user info. Otherwise this looks good to me

openedx_events/learning/data.py
ilee2u: Looking at the parameters of the [downstream functions](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/577699957/M6+Hackathon+Outline#(Old)-Parameters-of-downstream-functions-in-edx-platform%3A), it looks like most of them take user_id as a parameter anyways, so I think user_id is the way to go here.

Also I believe Michael's PR just has placeholder data for now.

MichaelRoytman: I think it would make sense to use the `UserData`. It seems to be more common to use that than a `user_id`. But I do see uses of both, so there's precedence to use both.

I don't think we have to model our events off the way they're being consumed now. Ideally, they shouldn't be so context aware. I think, for me, it's less, "the Instructor Service currently only expects a `user_id` so that's what we should put in the event" and more "how can I best and most generically represent the concept of a user in my event"?

I guess the only consideration is maybe sending PII when we don't need to?

openedx_events/learning/signals.py
```diff
@@ -196,3 +197,58 @@
...
 )
+
+# .. event_type: org.openedx.learning.exam.attempt.submitted.v1
+# .. event_name: EXAM_ATTEMPT_SUBMITTED
+# .. event_description: Emitted when an exam attempt is submitted by a learner.
+# .. event_data: ExamAttemptData
+EXAM_ATTEMPT_SUBMITTED = OpenEdxPublicSignal(
+    event_type="org.openedx.learning.exam.attempt.submitted.v1",
+    data={
+        "exam_attempt_data": ExamAttemptData,
```
MichaelRoytman, line 208: Nit. I think `exam_attempt_data` could be a little verbose and `data` is a little redundant. What do you think about just `exam_attempt`? That matches the other events more closely.

ilee2u: Yeah that makes sense, changing now

openedx_events/learning/data.py
```diff
@@ -296,3 +296,23 @@ class ProgramCertificateData:
...
+    course_key = attr.ib(type=CourseKey)
+    usage_key = attr.ib(type=UsageKey)
+    requesting_user = attr.ib(type=UserData, default=None)
+    credit_requirement_status = attr.ib(type=str, default=None)
```
MichaelRoytman, line 318: I thought we didn't need this anymore. Did something change?

ilee2u: Nope just forgot to remove it

openedx_events/learning/data.py
Zacharis278: I think there's a good chance future downstream logic would use the pii. This probably gets hooked up to name affirmation at some point.

Zacharis278 (approved):

MichaelRoytman (approved):
Nice job!

openedx_events/learning/data.py
```diff
@@ -296,3 +296,22 @@ class ProgramCertificateData:
...
+class ExamAttemptData:
+    """
+    Attributes defined for the Open edX Exam downstream effects.
+
+    Arguments:
+        student_user (UserData): User object for the student to whom the exam attempt belongs
+        course_key (CourseKey): identifier of the course to which the exam attempt belongs
+        usage_key (UsageKey): identifier of the content that will get a exam attempt
+        requesting_user (UserData): user triggering the event (sometimes a non-learner, e.g. an instructor)
+        credit_requirement_status (str): what to change the student's CreditRequirement status to
```
robrap, line 311: Docstring vs implementation for credit_requirement_status?

ilee2u: Whoops! Fixing that now.

ilee2u: For context, I removed that parameter and just forgot to remove the docstring.

openedx_events/learning/signals.py
```diff
@@ -196,3 +197,58 @@
         "notification_data": UserNotificationData,
     }
 )
+
+# .. event_type: org.openedx.learning.exam.attempt.submitted.v1
+# .. event_name: EXAM_ATTEMPT_SUBMITTED
+# .. event_description: Emitted when an exam attempt is submitted by a learner.
```
robrap, line 203: Some high-level questions:
1. Does the platform have other types of "exams" that would not trigger these events? If so, how do we differentiate? Note that an event is defined by its event trigger + event data.
2. For the exams covered by these events, are there multiple exam types? Would it be useful to know that as part of the event data?
3. If there are different exam types, do some of these events only apply to certain types of exam types? If so, would that be useful to note?

Zacharis278: 1. kind of. The existing exam system (edx-proctoring) does not trigger these events, the service that emits these is intended to replacement. There are no other types of exams outside that.
2. Good point. There are multiple exam types and edx-platform is _mostly_ unaware of this since all the state/ui flows are manage by the exams/proctoring service. That said, I think there is value including this. Consumers may likely care if an exam is proctored or not. We might have actually missed a need for this. I think Credits depend on if an exam is proctored.
3. Yup some of these events are impossible depending on the exam type. A timed exam cannot be rejected since there is no video review to reject for example. We should add those details to the event descriptions

robrap: Are there non-timed exams, and are those included in this as well? Or are these just timed-exams, proctored and not-proctored? Just curious.

Zacharis278: 🤔  I guess this doesn't include 'all' exams just 'special exams' which have a different UX and workflow. Right now, all special exam types are a subset of timed. This wouldn't include normal sections instructors are using as or are graded as exams however. We could possibly reintroduce the term 'special exam' for these events which could clear this up.

robrap: If you think of this more as an API discussion around exams and special-exams, who do you think should be a part of this discussion? Maybe https://openedx.atlassian.net/wiki/spaces/COMM/pages/3554214289/Arch+Coordination+-+How+to+announce+and+tune-in will help you get the right discussion going?

robrap: Related: There are suggestions https://docs.openedx.org/projects/openedx-events/en/latest/how-tos/creating-new-events.html#propose-the-new-event-to-the-community for how to propose these new events.

ilee2u: Hi Robert!

To answer your question, you could technically title a course subsection as an "Exam" and grade it as an exam, however our exams backend (i.e. the exams IDA we're developing, edx-exams) would never know about it, as the exam backend is only used/called when a subsection is marked as a "Special Exam", i.e. a timed or proctored exam.

As such, the events that we are defining would only ever be produced in the context of a timed or proctored exam. The logic that we're implementing to produce these events mirrors the logic that edx-proctoring uses to call functions in various edx-platform services (e.g. calling the credit service to set a leaner's credit requirement status).

The bulk of the logic that we're porting from edx-proctoring can be seen here: https://github.com/openedx/edx-proctoring/blob/8091977772ec8967c79e854de7e0c813ee7c3d7e/edx_proctoring/api.py#L1479 (I recommend viewing this in an IDE since it can go a little deep).

Since certain exam attempt statuses (i.e. verified, rejected, and errored) are inherently linked to proctored exams, this means that edx-proctoring can distinguish if an exam is proctored or not based on the incoming status to update the attempt to (Side note that we never have a case where we need to know if an exam is specifically timed and not proctored).

Since we are simply mirroring this behavior in the exams IDA, this means that events produced in the exams IDA would be consumed by the intended services naturally, without us needing to gate anything behind a if exam_type=='proctored' kind of statement.

The pseudocode of how we'll mimic edx-proctoring's logic in edx-exams can be seen in an internal document that I can DM to you.

With all this being said, I agree that there could be value in attaching the exam type to the event down the line.

robrap: > We could possibly reintroduce the term 'special exam' for these events which could clear this up.

Something like this sounds reasonable, but the point of my last two messages is that it makes sense to bring in some other domain experts to help decide this type of question.

robrap: @ilee2u: I posted my last comment at the same time you were posting yours, but it applies equally to you. This all sounds reasonable, but it would be great to get others involved who work in this area to help with names and details to make the events more clear and forward thinking.

Zacharis278: @ilee2u I think a discuss post referencing this PR in combination with the [original ADR](https://github.com/openedx/edx-proctoring/pull/1049) for the project would be a good start to get input

ilee2u: A Discuss post for this topic has been created here: https://discuss.openedx.org/t/how-to-implement-events-for-the-exams-ida/11276

robrap: Can you confirm that you agreed/decided that no name or doc changes were needed to clarify what is meant by "exam"? It feels like at least a doc update might be warranted, unless you think I'm the only person that would ever ask about this.

mariajgrimaldi: Hello folks. I do believe we should use a different name for the event. `exam` might mean something in `edx-exams`, but it could mean other (similar) things in `edx-platform` that need another type of data for the event. What do you think?

Zacharis278: ^ I think this is a good idea to have before we merge. I could see people getting confused on two points

1. these events only pertain to 'special' exams
2. these events are not published by the existing exams system

openedx_events/learning/data.py
```diff
@@ -296,3 +296,23 @@ class ProgramCertificateData:
...
+class ExamAttemptData:
+    """
+    Attributes defined for the Open edX Exam downstream effects.
+
+    Arguments:
+        student_user (UserData): user object for the student to whom the exam attempt belongs
+        course_key (CourseKey): identifier of the course to which the exam attempt belongs
+        usage_key (UsageKey): identifier of the content that will get a exam attempt
+        requesting_user (UserData): user triggering the event (sometimes a non-learner, e.g. an instructor)
+        exam_type (str): type of exam that was taken (e.g. timed, proctored, etc.)
```
mariajgrimaldi, line 311: Nit:
```suggestion
        student_user (UserData): user object for the student to whom the exam attempt belongs
        course_key (CourseKey): identifier of the course to which the exam attempt belongs
        usage_key (UsageKey): identifier of the content that will get a exam attempt
        exam_type (str): type of exam that was taken (e.g. timed, proctored, etc.)
        requesting_user (UserData): user triggering the event (sometimes a non-learner, e.g. an instructor)
```

mariajgrimaldi (commented):
Hi there folks. I left a few minor comments. Thanks!

openedx_events/learning/signals.py
ilee2u: I've made the following changes to address this:

- Changed the data name to "SpecialExamAttemptData"
- Added a comment about how events using this data are only emitted for special exams from the newer exams backend, edx-exams

mariajgrimaldi: That's better! Thank you folks.

CHANGELOG.rst
```diff
@@ -17,6 +17,12 @@ Changed
 ~~~~~~~
 * Re-licensed this repository from AGPL 3.0 to Apache 2.0
 
+[8.670] - 2023-09-21
```
mariajgrimaldi, line 20:
```suggestion
[8.7.0] - 2023-09-29
```

ilee2u: Done 👍

ilee2u:
Hi, we've been informed that v8.6.0, which this PR is based off, is having issues, so please do not merge this until we find a solution/workaround for those issues.

openedx_events/learning/signals.py
robrap: I don't want to push hard on this, because it's not my domain of expertise, but the naming difference between EXAM_ATTEMPT_SUBMITTED and SpecialExamAttemptData just catches my attention. The obvious other option is to rename to SPECIAL_EXAM_ATTEMPT_SUBMITTED , and I guess the question I have is around the pros/cons of going with the more specific name.

openedx_events/learning/data.py
```diff
@@ -296,3 +296,29 @@ class ProgramCertificateData:
...
     url = attr.ib(type=str)
     certificate_available_date = attr.ib(type=datetime, default=None)
+
+
+@attr.s(frozen=True)
+class SpecialExamAttemptData:
+    """
+    Attributes defined for the Open edX Exam downstream effects.
+
+    Note that events that use this data type:
```
robrap, line 306: Do we want another point, like?:
```
* Use the prefix `EXAM_`, which is equivalent to "special exam", but the shorter name is to hopefully avoid confusion because these are the events and exams that would likely be of concern.
```
Feel free to reword.

openedx_events/learning/signals.py
Zacharis278: 🤔 🤔 It's a good question, I'd almost lean towards not using 'special' in the name at all and just commenting to the meaning? Otherwise, we need to consistently use the term across Open edX. I don't think 'special exam' has any additional meaning to anyone working outside this domain unless it's described. There was a actually discussion a while back that we stop using this term because it's just confusing, trying to dig that up.

I'm not sure how to best describe the fact that this intends to include all content that is treated with an exam user experience, but not content that might be titled (on the section or grading policy) by course teams as 'exam'.  The latter is ultimately no different from other content, the platform doesn't do anything functionally different. (hence not special)

docs/decisions/0013-special-exam-submission-and-review-events.rst
```diff
@@ -0,0 +1,37 @@
...
+*******
+
+* A new backend for exams called `edx-exams` is being developed (See the `exams IDA ADR <https://github.com/openedx/edx-proctoring/blob/master/docs/decisions/0004-exam-ida.rst>`_ for more info).
+* We are currently working to implement the downstream effects triggered whenever an exam attempt is submitted or reviewed. (For example, when an exam attempt is submitted, we will want to make sure `edx-platform` knows to mark the exam subsection as completed.)
+
+
+Decision
+********
+
+* Since `edx-exams` is independent, we will want to prevent circular dependencies between `edx-platform` and `edx-exams`.
```
Zacharis278, line 19: I think this isn't all that important from the perspective of someone working with events.

docs/decisions/0013-special-exam-submission-and-review-events.rst
```diff
@@ -0,0 +1,37 @@
+12. Event definitions for special exam post-submission and review
+#################################################################
+
+Status
+******
+
+**Draft** 2023-09-29
+
+Context
```
Zacharis278, line 9: I think the context section might be a good place to explain the meaning of exam / special exam.

Rough points to hit
- course sections with an `exam_type` have additional logic that governs completion, grading, credit requirements, and more based on the type of exam.
- these sections are sometimes referred to as special exams
- course sections without an exam type configured may still be graded as an exam. This type of content does not have the exam user experience and is not governed by any exam specific logic.

docs/decisions/0013-special-exam-submission-and-review-events.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Decision
+********
+
+Where these events will be produced/consumed:
+=============================================
+
+* `edx-exams` will produce these events.
+    * NOTE: There is no plan to have the legacy exams backend, `edx-proctoring`, produce these events.
+* `edx-platform` will consumed these events in order to handle all behavior as it pertains to the state of an exam subsection.
```
Zacharis278, line 34: nit: 
```suggestion
* `edx-platform` will consume these events in order to handle all behavior as it pertains to the state of an exam subsection.
```

mariajgrimaldi:
Hi @ilee2u! This PR introduced an ADR with the status draft. Do you think we can change the ADR's status to accepted by now?

ilee2u:
> > Hi @ilee2u! This PR introduced an ADR with the status draft. Do you think we can change the ADR's status to accepted by now?
> 
> Hi again, yes we should be able to! Thanks :)

Just made a PR: https://github.com/openedx/openedx-events/pull/410

PR #305, opened 2024-01-24
feat: use new setting to set ce_source

rgraber:
**Description:** Use a new EVENTS_SERVICE_NAME setting to determine the source in EventsMetadata. Defaults to the old SERVICE_VARIANT setting if not set. SERVICE_VARIANT is usually only set on lms or cms, and has other uses, so it makes sense to have a new variable that all services can use for the events.

I didn't talk about how the EVENTS_SERVICE_NAME will be used to set the client id because that is Kafka specific.

**Testing instructions:**
pip install -e openedx-events
set EVENTS_SERVICE_NAME in your service settings
fire an event
check that the event source is openedx/<EVENTS_SERVICE_NAME>/web

**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Unreleased
 --------------------
 Changed
 ~~~~~~~
-* Choose new EVENT_BUS_APP_NAME setting over SERVICE_VARIANT for data source
+* Allow new EVENT_BUS_APP_NAME setting to override SERVICE_VARIANT for data source.
 
 [9.2.0] - 2023-11-16
 --------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Unreleased
 --------------------
 Changed
 ~~~~~~~
-* Allow new EVENTs_SERVICE_NAME setting to override SERVICE_VARIANT for data source.
+* Allow new EVENTS_SERVICE_NAME setting to override SERVICE_VARIANT for data source.
 
 [9.2.0] - 2023-11-16
 --------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Unreleased
 --------------------
 Changed
 ~~~~~~~
-* Allow new EVENT_BUS_APP_NAME setting to override SERVICE_VARIANT for data source.
+* Allow new EVENTs_SERVICE_NAME setting to override SERVICE_VARIANT for data source.
 
 [9.2.0] - 2023-11-16
 --------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ----------
 
+[9.3.0] - 2024-01-24
+--------------------
+Changed
+~~~~~~~
+* Choose new EVENT_BUS_APP_NAME setting over SERVICE_VARIANT for data source
+
 [9.2.0] - 2023-11-16
 --------------------
 Added
--- a/openedx_events/data.py
+++ b/openedx_events/data.py
@@ -29,26 +29,23 @@ def _ensure_utc_time(_, attribute, value):
 
 def get_service_name():
     """
-    Get the service name that should be used in the source of an event.
+    Get the service name of the producing/consuming service of an event (or None if not set).
     """
     # .. setting_name: EVENTS_SERVICE_NAME
     # .. setting_default: None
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ----------
 
+[9.3.0] - 2024-01-24
+--------------------
+Changed
+~~~~~~~
+* Choose new EVENT_BUS_APP_NAME setting over SERVICE_VARIANT for data source
```
robrap, line 21: I thought you were replacing. How about something like?
```suggestion
* Allow new EVENT_BUS_APP_NAME setting to override SERVICE_VARIANT for data source.
```

openedx_events/data.py
```diff
@@ -27,6 +27,16 @@ def _ensure_utc_time(_, attribute, value):
     raise ValueError(f"'{attribute.name}' must have timezone.utc")
 
 
+def get_source_name():
+    """
+    Get the value that should be used for the source of an event.
+    """
+    source = getattr(settings, "EVENT_BUS_SERVICE_NAME", None)
```
robrap, line 34: 1. I think we need setting annotations here.
2. Maybe just on the PR, and maybe in the docs, you should explain:
  - Why we need the new setting, and when we should use it over the SERVICE_VARIANT setting. I imagine you basically wish to deprecate use of SERVICE_VARIANT here?
  - Is it ok that an "EVENT_BUS" setting will be used for all events, regardless of whether or not they are sent over the event bus? It _may_ be ok, but maybe we should note this.

Not sure if we need a small ADR around this, or if we can just use code comments?

rgraber: I didn't put settings annotations here because I put them in event-bus-kafka. Not sure what the protocol is when the same setting is used in 2 different places. I can change that.
I'd indeed like to deprecate use of service variant since that only really applies to lms and cms.
I'd be fine with EVENTS_SERVICE_NAME but we do need to decide on one that we'll use both here and in EBK.

rgraber: I really don't think this needs an ADR

robrap: [inform] @marcotuts @felipemontoya

robrap: @mariajgrimaldi: I don't know if this will notify you twice, but I added the wrong username and had to edit it above. I'm adding this in case you never got a notification.

rgraber: The PR description explains why we shouldn't use SERVICE_VARIANT. I think it would actually be more confusing to talk about it more in the code comments.

openedx_events/data.py
```diff
@@ -62,7 +75,8 @@ class EventsMetadata:
     source = attr.ib(
         type=str, default=None,
         converter=attr.converters.default_if_none(
-            attr.Factory(lambda: "openedx/{service}/web".format(service=getattr(settings, "SERVICE_VARIANT", "")))
+            attr.Factory(lambda: "openedx/{service}/web".format(service=get_source_name()) if get_source_name()
```
robrap, line 78: NIt: 
1. I'd find this easier to read if on and earlier line you set:
```
service_name = "unidentified" or _get_service_name().  # see other PR comment for method name change
```
2. Consider something like `SERVICE-NAME-UNSET` in place of `unidentified`, to give something easy to search for and easy to resolve.

openedx_events/data.py
```diff
@@ -27,6 +27,19 @@ def _ensure_utc_time(_, attribute, value):
     raise ValueError(f"'{attribute.name}' must have timezone.utc")
 
 
+def get_source_name():
```
robrap, line 30: 1. Should be `get_service_name`.
2. Can we keep this internal-only and call it `_get_service_name`?

robrap: Also note that if this function were private, and just used for this one purpose, it could return `SERVICE-NAME-UNSET` itself in place of None.

rgraber: I wanted to call it from event bus kafka

robrap: It should still be changed to `get_service_name` though, right?

rgraber: It was. I have 2 methods now

robrap (approved):

PR #309, opened 2024-01-29
feat: add courseaccessrole events

Zacharis278:
**Description:** Creates new data class and events for publishing information about the creation or removal of an LMS [CourseAccessRole](https://github.com/openedx/edx-platform/blob/96b360a1ee9c756b6bd1032dc282bdb24d3f152c/common/djangoapps/student/models/user.py#L1044). We are not planning to account for the [roles and permissions changes](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3824648277/RBAC+Tech+Spec) at this point, events would be based on how roles currently work.

This will be consumed by [edx-exams](https://github.com/edx/edx-exams) to sync access to instructor tools and actions on the API.

**Author concerns:** These events will be used instead those added by https://github.com/openedx/openedx-events/pull/267. I plan to remove those in a separate PR since there is no plan to use those event definitions or supporting data classes.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/openedx_events/learning/data.py
+++ b/openedx_events/learning/data.py
@@ -332,6 +332,24 @@ class ExamAttemptData:
 
 
 @attr.s(frozen=True)
+class CourseAccessRoleData:
+    """
+    Attributes defined for the Open edX Course Access Role data object.
+
+    Arguments:
+        user (UserData): user associated with the CourseAccessRole.
+        course_key (CourseKey): identifer of the related course object.
+        org (str): identifier of the organization.
+        role (str): the role of the user in the course.
+    """
+
+    user = attr.ib(type=UserData)
+    org_key = attr.ib(type=str)
+    course_key = attr.ib(type=CourseKey)
+    role = attr.ib(type=str)
+
+
+@attr.s(frozen=True)
 class ManageStudentsPermissionData:
     """
     Attributes defined for the Open edX to represent users that can manage students within a course/org.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,11 @@ Change Log
 
 Unreleased
 ----------
+[9.4.0] - 2024-01-29
+--------------------
+Added
+~~~~~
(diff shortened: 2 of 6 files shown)


openedx_events/learning/data.py
```diff
@@ -331,6 +331,24 @@ class ExamAttemptData:
     requesting_user = attr.ib(type=UserData, default=None)
 
 
+@attr.s(frozen=True)
+class CourseAccessRoleData:
```
alangsto, line 335: Reading the `ManageStudentsPermissionData` class defined below makes me wonder if we should refrain from using the word "role" in this class name. I know you've talked to more people about the roles + permissions work though, so I trust your judgement!

Zacharis278: thanks. Yup the conclusion we came to was to define this as edx-platform exists today rather than trying to guess the future. I kept the name specific to the LMS model so it's use is clear instead of a ambiguous half-measure that might be reusable in the future.

There's no longer a near term path to the permission changes getting merged as we had originally wanted to account for. I'm not sure manage students will even end up being the correct permission when that the time does come to use permissions over roles.

alangsto: Got it, appreciate the explanation! This looks good then 👍

felipemontoya (approved):
This addition of the events seems reasonable and well structured to me.

The removal of the other events will probably create a breaking change, but that can be discussed in its own PR.

Zacharis278:
thanks @felipemontoya. I don't have write access to merge this. Is that something you can do or do I need to ping a specific maintainer?

mariajgrimaldi (approved):
Thank you folks! I'll be merging this now.

mariajgrimaldi:
Done: https://github.com/openedx/openedx-events/releases/tag/v9.4.0

Issue #79, opened 2022-07-20
[Consumer] Prevent event bus recursion


timmc-edx:
It is possible (though moderately unlikely) that an IDA could inadvertently be configured to both produce a signal to the event bus and also consume that signal _from_ the event bus. This would result in any such event being reproduced and circulated indefinitely on a tight loop, causing quadratically increasing load and eventually an outage.

There are a couple of ways we could avoid this:

1. When converting an event to a Django signal-send, mark it as having been consumed from the event bus, and do not allow such marked signal-sends to be produced back onto the event bus. This is the simplest option and would only require having a special keyword arg that won't collide with other kwargs on the signal-send.
    - We may eventually start attaching event envelope/header information to signal-sends (e.g. to preserve and convey timestamps) and the mere presence of this information would suffice as a "mark".
3. Have openedx-events refuse to configure both a consumer and a producer for the same _topic_ in the same IDA (raise exception on startup). This is a bigger hammer and may not be what we want.
4. Record the IDA-of-origin on any OpenEdxPublicSignal send and copy that onto events and signal-sends when producing and consuming, then in the consumer drop any event that originated on the IDA on a previous iteration. Much more complicated, but would handle the (less likely) case of mutual recursion between 2+ IDAs.

Any such safeguard would not be able to detect anything that explicitly receives, recreates, and re-sends a signal, but that sort of thing is probably not worth worrying about.

ormsbee:
Some random thoughts:

CloudEvents does have a [source](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0041-arch-async-server-event-messaging.html#id3) field that we could use to see where something came from.

That being said, I think that option (2) here is a nice, big, simple hammer and we should use it until it starts getting in someone's way and then re-evaluate at that time. If a service is catching an event that it emitted, it might as well use the signal version of it instead, right?

In general, I think we should try to make it so that events either have one producer and many receivers, or many producers and one receiver--partly so we can avoid cycle issues. I'm sure there are valid use cases outside of those two modes, but I feel like any such use cases should be strongly examined to check for alternatives.

ormsbee:
> I was somewhat partial to option 2, but one wrinkle is that in our current semi-prototype code we're making explicit producer and consumer calls, which would bypass the protection that that option contemplates. (It would only work if all "signal teleporting" is fully configuration-based, so that openedx-events would have complete visibility and control.)

[...]

PR #312, opened 2024-02-09
fix: avoid infinite recursion of openedx-event

Ian2012:
**Description:** This PR adds a new argument `from_event_bus` to the `send_event` method (and related ones) that allows marking a Django signal as already processed from the event bus and avoids infinite emission of the same signal if such signal is consumed in the same runtime.

The following code block must be used on the consumer to avoid processing the signal in the same process that emitted the signal:

```python
from openedx_events.tooling import SIGNAL_PROCESSED_FROM_EVENT_BUS

...

def my_signal(...):
    if not kwargs.get(SIGNAL_PROCESSED_FROM_EVENT_BUS, False):
        logger.info("Event received from a non-event bus backend, skipping...")
        return
```

This has been tested with the `event-bus-redis` without any changes as the behavior of calling `send_event_with_custom_metadata` is to enable the `from_event_bus` param. Tested on https://github.com/openedx/event-routing-backends/pull/361 with the signal emitted on https://github.com/openedx/event-tracking/pull/246

**ISSUE:** Solves: https://github.com/openedx/openedx-events/issues/79

**Testing instructions:**

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/openedx_events/apps.py
+++ b/openedx_events/apps.py
@@ -1,13 +1,16 @@
 """
 openedx_events Django application initialization.
 """
+import logging
 
 from django.apps import AppConfig
 from django.conf import settings
 
 from openedx_events.event_bus import get_producer
 from openedx_events.exceptions import ProducerConfigurationError
-from openedx_events.tooling import OpenEdxPublicSignal, load_all_signals
+from openedx_events.tooling import SIGNAL_PROCESSED_FROM_EVENT_BUS, OpenEdxPublicSignal, load_all_signals
+
+logger = logging.getLogger(__name__)
 
 
 def general_signal_handler(sender, signal, **kwargs):  # pylint: disable=unused-argument
@@ -20,6 +23,13 @@ def general_signal_handler(sender, signal, **kwargs):  # pylint: disable=unused-
     #        "topic_a": { "event_key_field": "my.key.field", "enabled": True },
     #        "topic_b": { "event_key_field": "my.key.field", "enabled": False }
     # }"
+    if kwargs.get(SIGNAL_PROCESSED_FROM_EVENT_BUS) is True:
+        logger.debug(
+            "Declining to send signal to the Event Bus since that's "
+            f"where it was sent from: {signal.event_type} (preventing recursion)"
+        )
+        return
+
     event_data = {key: kwargs.get(key) for key in signal.init_data}
 
     for topic in event_type_producer_configs.keys():
--- a/openedx_events/tests/test_producer_config.py
+++ b/openedx_events/tests/test_producer_config.py
@@ -55,6 +55,29 @@ class ProducerConfiguratonTest(TestCase):
             call_args
         )
 
+    @patch("openedx_events.apps.logger")
+    @patch('openedx_events.apps.get_producer')
+    def test_send_events_with_custom_metadata_not_replayed_by_handler(self, mock_producer, mock_logger):
+        """
+        Check wheter XBLOCK_PUBLISHED is connected to the handler and the handler
+        do not send any events as the signal is marked "from_event_bus".
+
+        Args:
+            mock_producer: mock get_producer to inspect the arguments.
+            mock_logger: mock logger to inspect the arguments.
+        """
+        mock_send = Mock()
+        mock_producer.return_value = mock_send
+        metadata = XBLOCK_PUBLISHED.generate_signal_metadata()
+
+        XBLOCK_PUBLISHED.send_event_with_custom_metadata(metadata, xblock_info=self.xblock_info)
+
+        mock_send.send.assert_not_called()
+        mock_logger.debug.assert_called_once_with(
+            "Declining to send signal to the Event Bus since that's "
+            f"where it was sent from: {XBLOCK_PUBLISHED.event_type} (preventing recursion)"
+        )
+
     @patch('openedx_events.apps.get_producer')
     @override_settings(EVENT_BUS_PRODUCER_CONFIG={})
     def test_events_not_in_config(self, mock_producer):
--- a/openedx_events/tooling.py
+++ b/openedx_events/tooling.py
@@ -30,6 +30,8 @@ KNOWN_UNSERIALIZABLE_SIGNALS = [
     "org.openedx.learning.course.notification.requested.v1",
 ]
 
+SIGNAL_PROCESSED_FROM_EVENT_BUS = "from_event_bus"
+
 
 class OpenEdxPublicSignal(Signal):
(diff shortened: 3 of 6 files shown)


openedx-webhooks:
Thanks for the pull request, @Ian2012! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

openedx_events/apps.py
```diff
@@ -20,6 +22,10 @@ def general_signal_handler(sender, signal, **kwargs):  # pylint: disable=unused-
     #        "topic_a": { "event_key_field": "my.key.field", "enabled": True },
     #        "topic_b": { "event_key_field": "my.key.field", "enabled": False }
     # }"
+    if SIGNAL_PROCESSED_FROM_EVENT_BUS in kwargs and kwargs[SIGNAL_PROCESSED_FROM_EVENT_BUS] is True:
+        logger.info(f"Signal already processed on the Event Bus {signal.event_type}, skipping...")
```
Ian2012, line 26: Not sure if this should be logged or if should be behind the `AUDIT_LOGGING` flag

mariajgrimaldi (commented):
I left a few no-blocking comments! Thank you

openedx_events/apps.py
```diff
@@ -20,6 +23,10 @@ def general_signal_handler(sender, signal, **kwargs):  # pylint: disable=unused-
     #        "topic_a": { "event_key_field": "my.key.field", "enabled": True },
     #        "topic_b": { "event_key_field": "my.key.field", "enabled": False }
     # }"
+    if SIGNAL_PROCESSED_FROM_EVENT_BUS in kwargs and kwargs[SIGNAL_PROCESSED_FROM_EVENT_BUS] is True:
```
mariajgrimaldi, line 26: Tests are failing because of coverage. Maybe we can test that signals are not sent when the SIGNAL_PROCESSED_FROM_EVENT_BUS is passed?

openedx_events/apps.py
Ian2012: Yes, it is useful for debugging but this change is also related to the event bus implementation

robrap: Would `logger.debug` be more appropriate? In certain cases, couldn't this be common and expected and just cause a lot of unnecessary logging?

mariajgrimaldi: Using logger.debug would be better than introducing a new flag in the repository. So operators can still access these types of logs but have to configure the Python logger to do so.

openedx_events/tests/test_producer_config.py
```diff
@@ -55,6 +55,26 @@ def test_enabled_disabled_events(self, mock_producer):
...
 
+    @patch("openedx_events.apps.logger")
+    @patch('openedx_events.apps.get_producer')
+    def test_events_with_custom_metadata_disabled(self, mock_producer, mock_logger):
+        """
+        Check whether XBLOCK_PUBLISHED is connected to the handler and the handler only produces enabled events.
+
+        Args:
+            mock_producer: mock get_producer to inspect the arguments.
+        """
```
mariajgrimaldi, line 66: I think this docs need to be updated

openedx_events/tests/test_producer_config.py
```diff
@@ -55,6 +55,26 @@ def test_enabled_disabled_events(self, mock_producer):
             call_args
         )
 
+    @patch("openedx_events.apps.logger")
+    @patch('openedx_events.apps.get_producer')
+    def test_events_with_custom_metadata_disabled(self, mock_producer, mock_logger):
```
mariajgrimaldi, line 60: but you're using custom metadata, how is it disabled?

mariajgrimaldi:
The next release with these changes will affect the event bus implementation tooling, so I'm pinging @robrap @timmc-edx so they know. Thanks!

openedx_events/tests/test_producer_config.py
Ian2012: Named updated to match what it does: signals are not sent by the handler on the consumer side if those are sent from the event bus.

Ian2012:
> The next release with these changes will affect the event bus implementation tooling, so I'm pinging @robrap @timmc-edx so they know. Thanks!

@mariajgrimaldi it will not affect them as `event-bus-redis` and `event-bus-kafka` already call `send_event_with_custom_metadata` on receiving signals from the event bus. There doesn't exist a signal that is produced/consumed on the same runtime (except this current work: https://github.com/openedx/event-routing-backends/pull/361) and this PR makes the default behavior to mark the signal when calling such a method.

openedx_events/tooling.py
```diff
@@ -229,12 +238,18 @@ def send_event_with_custom_metadata(
         Arguments:
             metadata (EventsMetadata): The metadata to be sent with the signal.
             send_robust (bool): Defaults to True. See Django signal docs.
+            from_event_bus (bool): Defaults to False. If True, the signal is
```
timmc-edx, line 241: The param defaults to true, here.

timmc-edx (commented):
A few notes. Also remember to update version and changelog.

Ian2012:
@timmc-edx rebased and bumped. Do you think is ready to be merged? cc @mariajgrimaldi

Ian2012:
@timmc-edx rebased and bumped. Do you think is ready to be merged? cc @mariajgrimaldi

openedx_events/apps.py
```diff
@@ -20,6 +23,13 @@ def general_signal_handler(sender, signal, **kwargs):  # pylint: disable=unused-
     #        "topic_a": { "event_key_field": "my.key.field", "enabled": True },
     #        "topic_b": { "event_key_field": "my.key.field", "enabled": False }
     # }"
+    if kwargs.get(SIGNAL_PROCESSED_FROM_EVENT_BUS) is True:
+        logger.debug(
+            "Declining to send signal to the Event Bus since that's"
```
timmc-edx, line 28: [nit] Missing space (same in unit test)
```suggestion
            "Declining to send signal to the Event Bus since that's "
```

timmc-edx (commented):
One minor issue, but looks great overall!

mariajgrimaldi (approved):
Looking good! Thanks. I'll merge after @Ian2012 addresses the last comment.

PR #314, opened 2024-02-13
feat: removed unused permission events

Zacharis278:
**Description:** These events were added in anticipation of the [roles and permissions changes](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3824648277/RBAC+Tech+Spec). That work is now on hold and we have instead opted to publish an event based on roles as they exist in edx-platform today. See https://github.com/openedx/openedx-events/pull/309. As written this event might not match the future state of that work if it is picked up again. I propose these definitions be removed to avoid confusion.

**Reviewers:**
- [ ] tag reviewer
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.


alangsto (approved):

openedx_events/__init__.py
```diff
@@ -5,4 +5,4 @@
 more information about the project.
 """
 
-__version__ = "9.5.1"
+__version__ = "9.5.2"
```
felipemontoya, line 8: Given that we are removing events that were public, and might be used by any other developer that imported this code, this should be a breaking change with a major version bump.

Naturally the events are not in use by edx-plaform right now. Where they ever used by that repo?

Zacharis278: Nope, I think this would be safe to merge w/o considering a major bump since the 'permission' models this event would be based on don't even exist. I can confidently say this has never been used. That said, I don't have strong options on if this bumps a major version or not. Just figured I'd point that out if we'd like to avoid going to v10.

In hindsight, we should not have included these until we were more certain the new permission model would get merged into edx-platform.

mariajgrimaldi:
I'll look into the docs failure so we can merge this soon.

openedx_events/__init__.py
felipemontoya: If this was never really releaseda anywhere and has never been used I'm good with not bumping for major.

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ----------
 
+[9.5.2] - 2024-02-13
+--------------------
+Removed
+~~~~~
```
mariajgrimaldi, line 20:
```suggestion
Removed
~~~~~~~
```

mariajgrimaldi (changes requested):
A tiny change so tests pass. Thanks!

mariajgrimaldi (approved):
This is looking good to me, especially if the events were ever used. Thank you for the cleanup!

PR #341, opened 2024-04-18
refactor: update ora submission data

BryanttV:
### Description
This PR updates the `ORASubmissionData` class to send the same data as the [legacy analytics event](https://github.com/openedx/edx-ora2/blob/master/openassessment/xblock/apis/submissions/submissions_actions.py#L162-L171). This will facilitate the transition from one event to another if necessary.

> [!NOTE]
> As this is an event that has not been used yet, there will be no problem in adding these new fields.

The new class attributes are:
- `id` -> `uuid` (Renamed to maintain consistency with the legacy event)
- `anonymous_user_id`
- `location`
- `attempt_number`
- `created_at`
- `submitted_at`
- `answer` (`ORASubmissionAnswer`)
    - `parts`
    - `file_keys`
    - `file_descriptions`
    - `file_names`
    - `file_sizes`
    - `file_urls`

The new fields that are not in the legacy event are:
- `anonymous_user_id`: Anonymous user ID of the user who submitted the ORA submission.
- `location`: Location of the ORA submission.

This new data is important as it allows us to have more complete and detailed information on ORA submission at the event.

### DIFF
--- a/openedx_events/learning/data.py
+++ b/openedx_events/learning/data.py
@@ -465,14 +465,49 @@ class ORAFileDownloadsData:
 
 
 @attr.s(frozen=True)
+class ORASubmissionAnswer:
+    """
+    Attributes defined to represent the answer submitted by the user in an ORA submission.
+
+    Arguments:
+        parts (List[dict]): List with the response text in the ORA submission.
+
+        The following attributes are used to represent the files submitted in the ORA submission:
+
+        file_keys (List[str]): List of file keys in the ORA submission.
+        file_descriptions (List[str]): List of file descriptions in the ORA submission.
+        file_names (List[str]): List of file names in the ORA submission.
+        file_sizes (List[int]): List of file sizes in the ORA submission.
+        file_urls (List[str]): List of file URLs in the ORA submission.
+    """
+
+    parts = attr.ib(type=List[dict], factory=list)
+    file_keys = attr.ib(type=List[str], factory=list)
+    file_descriptions = attr.ib(type=List[str], factory=list)
+    file_names = attr.ib(type=List[str], factory=list)
+    file_sizes = attr.ib(type=List[int], factory=list)
+    file_urls = attr.ib(type=List[str], factory=list)
+
+
+@attr.s(frozen=True)
 class ORASubmissionData:
     """
     Attributes defined to represent event when a user submits an ORA assignment.
 
     Arguments:
-        id (str): identifier of the ORA submission.
-        file_downloads (List[ORAFileDownloadsData]): list of related files in the ORA submission.
+        uuid (str): The UUID of the ORA submission.
+        anonymous_user_id (str): Optional. Anonymous user ID of the user who submitted the ORA submission.
+        location (str): Optional. Location of the ORA submission.
+        attempt_number (int): Attempt number of the ORA submission.
+        created_at (datetime): Date and time when the ORA submission was created.
+        submitted_at (datetime): Date and time when the ORA submission was submitted.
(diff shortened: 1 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @BryanttV! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

openedx_events/learning/data.py
```diff
@@ -470,9 +470,21 @@ class ORASubmissionData:
...
     """
 
-    id = attr.ib(type=str)
+    uuid = attr.ib(type=str)
+    anonymous_user_id = attr.ib(type=str)
+    location = attr.ib(type=str)
+    attempt_number = attr.ib(type=int)
+    created_at = attr.ib(type=datetime)
+    submitted_at = attr.ib(type=datetime)
+    answer = attr.ib(type=dict)
```
mariajgrimaldi, line 489: A few comments:
1. I have enough context to know we didn't use this event before, so we can update the data without much trouble. Could we add that to the PR cover letter?
2. Should all this fields be required?

BryanttV: > Should all this fields be required?

I would think that the `location` and `anonymous_user_id` fields could be optional. The `file_downloads` field is already optional by having a factory.
The other fields being part of the submission data I think they should be required, what do you think?

mariajgrimaldi: > I would think that the location and anonymous_user_id fields could be optional. 
Is there a scenario in which this will not be sent by the event, or where those fields are none/empty? That's a good way of knowing if they should be optional. I agree the rest of them should be required, although, is the answer always present?

BryanttV: > Is there a scenario in which this will not be sent by the event, or where those fields are none/empty?

In the submission these fields are not found, so they may not always be available.

> is the answer always present?

Yes, in a submission there will always be a dictionary with the user's response, either with the response text, or information from the attached files.

mariajgrimaldi: > In the submission these fields are not found, so they may not always be available.
But they should always be available when sending the event, correct? So they are always non-empty.

> Yes, in a submission there will always be a dictionary with the user's response, either with the response text, or information from the attached files.
So what's the difference between answer and https://github.com/openedx/openedx-events/pull/341/files#diff-5cf3a0b26961626ce777acf38a3cd803cb0ab982f8f35a35d078cb9c62438fa3R450-R464?

mariajgrimaldi: If answer and FileDownloads overlap, should we create a class for answer and include the FileDownloads there? So we don't send two fields with the same information. Correct me if I get it wrong.

BryanttV: I followed your suggestion and updated the data classes, could you check it?

mariajgrimaldi (approved):
Thank you for addressing all my comments! Could you update the lib versions so we can merge?

openedx-webhooks:
@BryanttV 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-events&entry.867055334=https://github.com/openedx/openedx-events/pull/341&entry.1484655318=https://github.com/BryanttV&entry.752974735=2024-04-18+14:25&entry.1917517419=2024-04-18+19:39&entry.2133058324=Yes) so we can improve your experience in the future.

PR #380, opened 2024-07-30
feat: Add enterprise signal for learner credit fulfillment revokation

pwnage101:
ENT-9213

### DIFF
--- a/openedx_events/enterprise/data.py
+++ b/openedx_events/enterprise/data.py
@@ -94,3 +94,71 @@ class LedgerTransaction(BaseLedgerTransaction):
     parent_content_key = attr.ib(type=str, default=None)
     fulfillment_identifier = attr.ib(type=str, default=None)
     reversal = attr.ib(type=LedgerTransactionReversal, default=None)
+
+
+@attr.s(frozen=True)
+class EnterpriseCustomerUser:
+    """
+    Attributes of an ``enterprise.EnterpriseCustomerUser`` record.
+    """
+
+    id = attr.ib(type=int)
+    created = attr.ib(type=datetime)
+    modified = attr.ib(type=datetime)
+    enterprise_customer_uuid = attr.ib(type=UUID)
+    user_id = attr.ib(type=int)
+    active = attr.ib(type=bool)
+    linked = attr.ib(type=bool)
+    is_relinkable = attr.ib(type=bool)
+    invite_key = attr.ib(type=UUID)
+    should_inactivate_other_customers = attr.ib(type=bool)
+
+
+@attr.s(frozen=True)
+class EnterpriseCourseEnrollment:
+    """
+    Attributes of an ``enterprise.EnterpriseCourseEnrollment`` record.
+    """
+
+    id = attr.ib(type=int)
+    created = attr.ib(type=datetime)
+    modified = attr.ib(type=datetime)
+    enterprise_customer_user = attr.ib(type=EnterpriseCustomerUser)
+    course_id = attr.ib(type=CourseKey)
+    saved_for_later = attr.ib(type=bool)
+    source_slug = attr.ib(type=str)
+    unenrolled = attr.ib(type=bool)
+    unenrolled_at = attr.ib(type=datetime)
+
+
+@attr.s(frozen=True)
+class BaseEnterpriseFulfillment:
+    """
+    Defines the common attributes of enterprise fulfillment classes, i.e. ``enterprise.EnterpriseFulfillmentSource``.
+    """
+
+    uuid = attr.ib(type=UUID)
+    created = attr.ib(type=datetime)
+    modified = attr.ib(type=datetime)
+    fulfillment_type = attr.ib(type=str)
+    enterprise_course_entitlement_uuid = attr.ib(type=UUID)
+    enterprise_course_enrollment = attr.ib(type=EnterpriseCourseEnrollment)
+    is_revoked = attr.ib(type=bool)
+
+
+@attr.s(frozen=True)
+class LearnerCreditEnterpriseCourseEnrollment(BaseEnterpriseFulfillment):
+    """
+    Attributes of an ``enterprise.LearnerCreditEnterpriseCourseEnrollment`` record.
+    """
+
+    transaction_id = attr.ib(type=UUID)
+
+
+@attr.s(frozen=True)
+class LicensedEnterpriseCourseEnrollment(BaseEnterpriseFulfillment):
+    """
+    Attributes of an ``enterprise.LicensedEnterpriseCourseEnrollment`` record.
+    """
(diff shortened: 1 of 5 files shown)


openedx_events/enterprise/data.py
```diff
@@ -94,3 +94,71 @@ class LedgerTransaction(BaseLedgerTransaction):
     parent_content_key = attr.ib(type=str, default=None)
     fulfillment_identifier = attr.ib(type=str, default=None)
     reversal = attr.ib(type=LedgerTransactionReversal, default=None)
+
+
+@attr.s(frozen=True)
+class EnterpriseCustomerUser:
+    """
+    Defines the common attributes of enterprise fulfillment classes.
```
adamstankiewicz, line 102: nit: looks like some copy/pasta in this docstring and `EnterpriseCourseEnrollment` below. Perhaps something like the following?

```suggestion
    Defines the attributes of enterprise customer users.
```

openedx_events/event_bus/avro/tests/schemas/org+openedx+enterprise+learner_credit_course_enrollment+revoked+v1_schema.avsc
```diff
@@ -0,0 +1,135 @@
+{
```
pwnage101, line 1: This file is generated via
```
python manage.py generate_avro_schemas org.openedx.enterprise.learner_credit_course_enrollment.revoked.v1
```

openedx_events/enterprise/data.py
```diff
@@ -94,3 +94,71 @@ class LedgerTransaction(BaseLedgerTransaction):
...
+    linked = attr.ib(type=bool)
+    is_relinkable = attr.ib(type=bool)
+    invite_key = attr.ib(type=UUID)
+    should_inactivate_other_customers = attr.ib(type=bool)
+
+
+@attr.s(frozen=True)
+class EnterpriseCourseEnrollment:
+    """
+    Defines the common attributes of enterprise fulfillment classes.
```
adamstankiewicz, line 120: nit: similar as above nit, i believe this might be copy/pasta from the class below? perhaps something like the following?

```suggestion
    Defines the attributes of enterprise course enrollments.
```

openedx_events/enterprise/data.py
```diff
@@ -94,3 +94,71 @@ class LedgerTransaction(BaseLedgerTransaction):
...
+    """
+
+    id = attr.ib(type=int)
+    created = attr.ib(type=datetime)
+    modified = attr.ib(type=datetime)
+    enterprise_customer_user = attr.ib(type=EnterpriseCustomerUser)
+    course_id = attr.ib(type=CourseKey)
```
adamstankiewicz, line 127: [sanity check] for my own edification, does `CourseKey` adequately handle course _run_ keys?

adamstankiewicz (approved):
Looks good to me, with a couple docs-related nits and a sanity check type question.

openedx_events/enterprise/data.py
pwnage101: thanks, good catch.

openedx_events/enterprise/data.py
pwnage101: Yes, this is an age old question, but my understanding is that CourseKey from opaque-keys is supposed to represent a course run, and AFAIK it can't even accurately represent a course key sans run.

adamstankiewicz (approved):

pwnage101:
@mariajgrimaldi please take a look when you have a chance!

macdiesel (approved):

pwnage101:
Fair, so first here's the ticket body for ENT-9213:

> Background
> 
> There is an  enterprise-subsidy cron job that writes reversals from recently unenrolled records to get transactions into a reversed state (at which point we'll send an event that causes the corresponding assignment to be reversed).  We should eliminate that cron/mgmt command and replace it with an event producer/consumer.
> 
> AC
> 
> * Add a new event for enterprise unenrollment events `LearnerCreditEnterpriseCourseEnrollment`
>   * In particular, we want to know when an `EnterpriseCourseEnrollment` becomes unenrolled, and then send data about that enrollment and its related fulfillment record onto the event bus.
>   * Ensure transaction ID is in the message
>   * https://github.com/openedx/edx-enterprise/blob/d7e67a82fad599509ecd06a5ef7b52bbc8e3fe29/enterprise/models.py#L2325C7-L2325C46
> * enterprise-subsidy consumes these events in real time and writes reversal for that one learner.
>   * Ensure that enterprise-subsidy has appropriate roles, ACLs, API keys etc. to be a consumer of openedx events.

pwnage101:
Here's a list of all other openedx PRs which are all part of this same ticket:

* [feat\: emit learner credit unenrollment event · openedx\/edx\-enterprise](https://github.com/openedx/edx-enterprise/pull/2190)
* [feat\: install and enable new LC enrollment revoked event · openedx\/edx\-platform](https://github.com/openedx/edx-platform/pull/35218)
* [feat\: write reversal on LC enrollment revoked event · openedx\/enterprise\-subsidy](https://github.com/openedx/enterprise-subsidy/pull/282)

mariajgrimaldi:
Hi folks! Thank you for your contribution. I'll be reviewing this as soon as I can. Thanks for the patience!

pwnage101:
FYI, I also updated the docstrings with direct links to django models in openedx code.

mariajgrimaldi (approved):
A few non-blocking comments. Thanks!

openedx_events/enterprise/data.py
```diff
@@ -94,3 +94,80 @@ class LedgerTransaction(BaseLedgerTransaction):
...
+    id = attr.ib(type=int)
+    created = attr.ib(type=datetime)
+    modified = attr.ib(type=datetime)
+    enterprise_customer_uuid = attr.ib(type=UUID)
+    user_id = attr.ib(type=int)
+    active = attr.ib(type=bool)
+    linked = attr.ib(type=bool)
+    is_relinkable = attr.ib(type=bool)
+    invite_key = attr.ib(type=UUID)
+    should_inactivate_other_customers = attr.ib(type=bool)
```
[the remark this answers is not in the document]
pwnage101: The `id` corresponds to the PK of this record, while the `user_id` represents the LMS user ID. We have since adopted a pattern of naming it `lms_user_id`, but this django model was created long ago. I guess the event can name this field anything and doesn't have to match the model.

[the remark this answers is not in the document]
pwnage101: I think I'll just keep it matching the model.

pwnage101:
@mariajgrimaldi please merge at your earliest convenience, as I do not have write access to this repository.

PR #394, opened 2024-09-11
fix: make nullable django fields also Noneable in event data

pwnage101:
This should hopefully fix the error we are currently seeing when attempting to serialize real-life LC fulfillments outside of unit test environments.  The error is exactly described in this ticket for a different event experiencing the same issue:

https://github.com/edx/edx-arch-experiments/issues/788

Basically, certain fields are passed into the serializer as None because the actual value in the database is NULL, but the serializer field default is not explicitly set to None which triggers a code path that doesn't tolerate being passed None:

https://github.com/openedx/openedx-events/blob/3170416c157df480b3d9f9f276ab7cf7b9d5e68d/openedx_events/event_bus/avro/serializer.py#L34-L44

ENT-9213

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,14 @@ __________
 
 
 
+[9.14.1] - 2024-09-17
+---------------------
+
+Changed
+~~~~~~~
+
+* Fixed event ``LEARNER_CREDIT_COURSE_ENROLLMENT_REVOKED`` so that its serializer can tolerate null nullable fields.
+
 [9.14.0] - 2024-09-12
 ---------------------
 
--- a/openedx_events/__init__.py
+++ b/openedx_events/__init__.py
@@ -5,4 +5,4 @@ These definitions are part of the Hooks Extension Framework, see OEP-50 for
 more information about the project.
 """
 
-__version__ = "9.14.0"
+__version__ = "9.14.1"
--- a/openedx_events/enterprise/data.py
+++ b/openedx_events/enterprise/data.py
@@ -125,8 +125,8 @@ class EnterpriseCustomerUser:
     active = attr.ib(type=bool)
     linked = attr.ib(type=bool)
     is_relinkable = attr.ib(type=bool)
-    invite_key = attr.ib(type=UUID)
     should_inactivate_other_customers = attr.ib(type=bool)
+    invite_key = attr.ib(type=UUID, default=None)
 
 
 @attr.s(frozen=True)
@@ -154,9 +154,9 @@ class EnterpriseCourseEnrollment:
     enterprise_customer_user = attr.ib(type=EnterpriseCustomerUser)
     course_id = attr.ib(type=CourseKey)
     saved_for_later = attr.ib(type=bool)
-    source_slug = attr.ib(type=str)
-    unenrolled = attr.ib(type=bool)
-    unenrolled_at = attr.ib(type=datetime)
+    source_slug = attr.ib(type=str, default=None)
+    unenrolled = attr.ib(type=bool, default=None)
+    unenrolled_at = attr.ib(type=datetime, default=None)
 
 
 @attr.s(frozen=True)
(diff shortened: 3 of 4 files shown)


pwnage101:
@mariajgrimaldi Please take a look when you have a chance!  Upon testing, I learned that in order to serialize a None value in a field, the field default first needs to be set to None, or else the serializer errors out.

mariajgrimaldi (approved):
This looks good. Let's hope no other events raise the same errors. I'll make a mental note to look for these edge cases when reviewing. Thank you so much!

pwnage101:
Thanks for the prompt review!  I do want to get this right so now that I think about it, I want to perform additional testing.  I was just eager to get this out for review in case of delay, but that was not a concern.

Maybe you can also help answer a question about versioning: Since this is one of the first instances where we're "evolving" an existing event without creating a v2 version of it, do you forsee any possible issues of doing that?  Technical issues maybe? This change is theoretically backwards-compatible (only making existing fields more permissive), and I also doubt anybody in the open source community has adopted this event given it's limited use for enterprise purposes and age.

mariajgrimaldi:
@pwnage101: That's a good question. As you said, I see these changes as backward compatible too. If these events were adopted elsewhere, defaults must've been handled in some way for them to work, so these changes won't affect current implementations. In this case, documentation will be enough to let adopters know. However, if there was a default for the data classes, we'll have to discuss the implications further.

pwnage101:
I just finished testing locally, and indeed it works the way I expected.  First here's a reproduction of the error before any changes (`data` is [prepared in a unit test](https://github.com/openedx/edx-enterprise/blob/fa2a4e918e6f74a6218812905e6538276b564d03/tests/test_enterprise/test_event_bus.py#L42) to simulate what we'd pass to send_event()):

```
(Pdb) from edx_event_bus_kafka.internal.producer import get_signal_serializer
(Pdb) from openedx_events.enterprise.signals import LEARNER_CREDIT_COURSE_ENROLLMENT_REVOKED
(Pdb) serializer = get_signal_serializer(LEARNER_CREDIT_COURSE_ENROLLMENT_REVOKED)
(Pdb) serializer.to_dict({'learner_credit_course_enrollment': data})
*** Exception: None cannot be handled by custom serializers (and default=None was not set)
```

After changes to make fields allow None:

```
(Pdb) from edx_event_bus_kafka.internal.producer import get_signal_serializer
(Pdb) from openedx_events.enterprise.signals import LEARNER_CREDIT_COURSE_ENROLLMENT_REVOKED
(Pdb) serializer = get_signal_serializer(LEARNER_CREDIT_COURSE_ENROLLMENT_REVOKED)
(Pdb) serializer.to_dict({'learner_credit_course_enrollment': data})
{'learner_credit_course_enrollment': {'created': '2024-09-11T21:45:51.594917+00:00', 'enterprise_course_enrollment': {'course_id': 'week-beat-through', 'created': '2024-09-11T21:45:51.593861+00:00', 'enterprise_customer_user': {'active': True, 'created': '2024-09-11T21:45:51.586227+00:00', 'enterprise_customer_uuid': '9fb43fcb-a4de-4c60-a578-29b37a01b85b', 'id': 1, 'invite_key': None, 'is_relinkable': True, 'linked': True, 'modified': '2024-09-11T21:45:51.586227+00:00', 'should_inactivate_other_customers': True, 'user_id': 1}, 'id': 1, 'modified': '2024-09-11T21:45:51.593861+00:00', 'saved_for_later': False, 'source_slug': 'enterprise_api', 'unenrolled': None, 'unenrolled_at': None}, 'enterprise_course_entitlement_uuid': None, 'fulfillment_type': 'learner_credit', 'is_revoked': False, 'modified': '2024-09-11T21:45:51.594917+00:00', 'transaction_id': '9cffe896-f52c-4d31-89cb-d5d218261ca8', 'uuid': '38a66051-fcfe-4387-96d5-743c281670cf'}}
```

pwnage101:
I'm happy with this PR and feel it's ready to merge if you feel the same.

pwnage101:
@mariajgrimaldi please merge at your convenience

pwnage101:
@mariajgrimaldi could you take a look?

CHANGELOG.rst
```diff
@@ -18,6 +18,14 @@ __________
 
 
 
+[9.14.1] - 2024-09-12
```
mariajgrimaldi, line 21: Could you please update the date?

mariajgrimaldi (commented):
I'll be merging this today. Thank you!

PR #438, opened 2024-12-17
docs: [FC-0074] add ADR with event design suggestions

mariajgrimaldi:
## Description

This PR adds an ADR with suggestions on how to design a new event based on the [Building Event-Driven Microservices](https://www.oreilly.com/library/view/building-event-driven-microservices/9781492057888/) 3rd chapter about Communication and Data Contracts and Martin Fowler's [Event Driven article](https://martinfowler.com/articles/201701-event-driven.html). I'm proposing that these practices be considered when implementing new events, but they should NOT be considered standards that all events should strictly follow. Also, I'm not saying either that all the existing events in the library comply (or should) with these practices. 

This document will be used as a reference in the How to Create a New Event guide. I'm currently validating these items while I write the guide, making sure these suggestions are clear enough to follow. You can review it here: https://github.com/openedx/openedx-events/pull/439

## Testing instructions

Review here: https://docsopenedxorg--438.org.readthedocs.build/projects/openedx-events/en/438/decisions/0016-event-design-practices.html

## Deadline

None

## Checklists

Check off if complete *or* not applicable:

**Merge Checklist:**
- [ ] All reviewers approved
- [ ] Reviewer tested the code following the testing instructions
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added with short description of the change and current date
- [ ] Code dependencies reviewed
- [ ] Fixup commits are squashed away
- [ ] Unit tests added/updated
- [ ] Noted any: Concerns, dependencies, migration issues, deadlines, tickets

**Post Merge:**
- [ ] Create a tag
- [ ] Create a release on GitHub
      finished.
- [ ] Upgrade the package in the Open edX platform requirements (if applicable)


docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,33 @@
...
+-------
+
+It is important to follow standards to ensure that the events are consistent, maintainable, and reusable. The design of the events should be self-descriptive, self-contained, and provide enough information for consumers to understand the message. This ADR aims to provide a set of suggested practices for designing Open edX events that are consistent with the architecture and contribute to the overall quality of the Open edX ecosystem.
+
+Decision
+--------
+
+We have compiled a list of suggested practices taken from `Event-Driven Microservices`_ and the `Event-Driven article`_ that we recommend reviewing and following when designing an Open edX Event and contributing to the library. The goal is to implement events that are consistent with the architecture, reusable, and maintainable over time.
+
+#. An event should describe as accurately as possible what happened (what) with its and why it happened (why). It must contain enough information for consumers to understand the message.
```
mariajgrimaldi, line 19:
```suggestion
#. An event should describe as accurately as possible what happened (what) and why it happened (why). It must contain enough information for consumers to understand the message.
```

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,33 @@
...
+#. Design events to be self-descriptive and self-contained. It should contain all the information necessary about what took place for consumers to react to the event without consulting other data sources.
```
mariajgrimaldi, line 20:
```suggestion
#. Design events to be self-descriptive and self-contained. The event should contain all the information necessary about what took place for consumers to react to the event without consulting other data sources.
```

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,33 @@
...
+#. Avoid ambiguous data fields or fields with multiple meaning.
+#. Design events with a single responsibility in mind. Each event should represent a single action or fact.
```
mariajgrimaldi, line 22:
```suggestion
#. Design events with a single responsibility in mind. Each event should represent a single action or fact that happened in the system.
```

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,33 @@
...
+#. Avoid adding flow-control information or business logic to events. Events should be solely a representation of what happened.
```
mariajgrimaldi, line 23:
```suggestion
#. Avoid adding flow-control information or business logic to events. Events should be solely a representation of what took place.
```

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,33 @@
...
+#. Use appropriate data types and formats for the event fields.
+#. Design the event so it is small, well-defined and only contain relevant information. Avoid including unnecessary or unrelated context.
+#. Ensure the event carries all necessary data to prevent runtime dependencies with other services.
+#. Manage the granularity of the event so it is not too course (generic name with too much information) or too fine-grained (specific name with too little information). It should be expressive enough to be useful.
```
mariajgrimaldi, line 27:
```suggestion
#. Manage the granularity of the event so it is not too coarse (generic name with too much information) or too fine-grained (specific name with too little information). It should be expressive enough to be useful.
```

mariajgrimaldi:
FYI @Squirrel18

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Responsibility and Granularity
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Design events with a single responsibility in mind. Each event should represent a single action or fact that happened in the system. If an event contains multiple actions, consider splitting it into multiple events. For instance, if the course grade is updated to pass or fail, there should be two events: one for the pass action and another for the fail action.
+- Manage the granularity of the event so it is not too coarse (generic with too much information) or too fine-grained (specific with too little information). When making a decision on the granularity of the event, start with the minimum required information for consumers to react to the event and add more information as needed with enough justification. If necessary, leverage API calls from the consumer side to retrieve additional information but always consider the trade-offs of adding dependencies with other services.
```
Alec4r, line 37: I agree with the idea of avoid generic events, however, we should also avoid split events like course_passed or course_failed when we could use just course_completed with an status.

could we include a practical example on an appropiate level of granularity?

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,58 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Use appropriate data types and formats for the event fields. Don't use generic data types like strings for all fields. Use specific data types like integers, floats, dates, or custom types when necessary.
+- Avoid ambiguous data fields or fields with multiple meaning. For instance, if an event contains a field called ``status`` it should be clear what the status represents. If the status can have multiple meanings, consider splitting the event into multiple events or adding a new field to clarify the status.
+
+Consumer-Centric Design
+~~~~~~~~~~~~~~~~~~~~~~~
+
+- When designing an event, consider the consumers that will be using it. What information do they need to react to the event? What data is necessary for them to process the event?
+- Design events carefully from the start to minimize breaking changes for consumers, although it is not always possible to avoid breaking changes.
```
Alec4r, line 50: will it be necessary to version events to handle event changes, or what is the plan for handling event changes?

docs/decisions/0016-event-design-practices.rst
mariajgrimaldi: I'm interested in understanding why having a ``status`` is better. I think that in this case, it'd be easier to consume more granular events than leave the responsibility to the consumer to figure out whether the student passed or failed based on the status.

docs/decisions/0016-event-design-practices.rst
mariajgrimaldi: Events are versioned by definition, see this [ADR](https://github.com/openedx/openedx-events/blob/main/docs/decisions/0002-events-naming-and-versioning.rst#2-open-edx-events-naming-and-versioning) for more info. As for the evolution of the events schema, this other ADR describes what the behavior is supposed to be: https://github.com/openedx/openedx-events/blob/main/docs/decisions/0006-event-schema-serialization-and-evolution.rst#decision, although according to [this comment](https://github.com/edx/edx-arch-experiments/issues/53#issuecomment-1286003436) the ADR might be outdated -- I'll be working on updating it.  The reality is that we haven't needed to change an event definition in any way that's breaking.

mariajgrimaldi: @robrap: Can you help us figure out what needs to change from the ADR-0006? I could do it, but I need more context to do it effectively.

docs/decisions/0016-event-design-practices.rst
Alec4r: I understand your perspective, and I think both approaches have valid use cases depending on the context, let me elaborate on why I suggested using course_completed with a status field and we can discuss further to find a balance

If the event represents a single conceptual action for example: completing a course, having one event like course_completed with a clear status passed, failed, etc. could simplify the producer's logic and reduce event proliferation and for consumers, interpreting the status field is relatively straightforward if it's well-documented and includes only a few well-defined values, however if different consumers need to handle passed and failed cases in significantly different ways, separate those events might reduce complexity for them but emitting separate events like course_passed and course_failed could also create challenges such as needing to ensure mutual exclusivty.

So in cases like this I think We could start with course_completed and a status field, ensuring it is well-documented and strictly validted and if in the future we observe a clear need for distinct event flows We could introduce the more granular events without breaking existing consumers.

mariajgrimaldi: In the example we're using, I still think it'd be more straightforward to send smaller and more specific events. As I see it, course completion and grade passing would be two different critical facts happening in the system; therefore, they should be independent. This is more of what consumers would want with a course completion event or a course passing status change. 
In any case, these are the only suggestions that should be evaluated for each case. We currently have an event called `COURSE_PASSING_STATUS_UPDATED` which uses a similar flow control status field called `is_passing`. In that case, given that the status is directly related to the event, using a status field is acceptable. What do you think? Should we maybe use a rule of thumb indicating that an event can be split into more events, given that they could communicate more facts of the system?

mariajgrimaldi: I added some examples illustrating the granularity I'm referring to: https://github.com/openedx/openedx-events/pull/438/files#diff-bdc081c0ccc9885672f08f8b2c853ca7ce8a068db2cb2497d11e04b19013e19aR88-R112

sarina: @Alec4r I worry that your example is predicated on understanding the needs of the consumers. I don't think that's always possible, to predict how various consumers will need to consume events. I agree with @mariajgrimaldi that consuming an event that doesn't require me to _also_ check the status field is conceptually simpler, even if it means I need to handle consuming a few discrete events.

docs/decisions/0016-event-design-practices.rst
robrap: Hello @mariajgrimaldi. I was off for the last two weeks, so just getting back to things.

It seems like you are pretty up to date. I'll summarize what I think you are already saying.
1. As noted in the comment you linked to, the existence of optional fields probably changes the 0006 recommendation. The first sentence of the [deferred alternatives in the ADR](https://github.com/openedx/openedx-events/blob/main/docs/decisions/0006-event-schema-serialization-and-evolution.rst#deferredrejected-decisions) reads: "We could add the ability to add optional fields to an event...". And we did add this ability.
2. Most importantly, no events have actually gone through any process of evolution, so all of this is just best-guesses about what we'd like to do and how we think it would work. It was deemed too much effort to proceed any further without an actual need.

Does this help? Let me know if you have any specific questions.

efortish (commented):
The event design suggestions are pretty easy to understand, I think the main goal of this document is to provide clear information on how an event should be design and this document accomplishes that goal, even for someone who has no experience in this topic.

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,126 @@
...
+- Manage the granularity of the event so it is not too coarse (generic with too much information) or too fine-grained (specific with too little information). When making a decision on the granularity of the event, start with the minimum required information for consumers to react to the event and add more information as needed with enough justification. If necessary, leverage API calls from the consumer side to retrieve additional information but always consider the trade-offs of adding dependencies with other services.
+- Ensure that the triggering logic is consistent and narrow. For instance, if an event is triggered when a user enrolls in a course, it should be trigger only when the user enrolls in a course in all ways possible to enroll in a course. If the event is triggered when a user enrolls in a course through the API, it should also be triggered when the user enrolls in a course through the UI.
```
sarina, line 60:
```suggestion
- Ensure that the triggering logic is consistent and narrow. For instance, if an event is triggered when a user enrolls in a course, it should be triggered when the user enrolls in a course in all ways possible to enroll in a course. If the event is triggered when a user enrolls in a course through the API, it should also be triggered when the user enrolls in a course through the UI.
```

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,126 @@
...
+- The ``is_active`` field is a boolean that represents whether the enrollment is active or not.
+- The ``creation_date`` field is a datetime that represents the creation date of the enrollment.
+- The ``created_by`` field is a ``UserData`` that represents the user who created the enrollment.
+- The ``user`` field is a ``UserData`` that represents the user associated with the Course Enrollment.
+- The ``course`` field is a ``CourseData`` that represents the course where the user is enrolled in.
+
+Consumer-Centric Design
+~~~~~~~~~~~~~~~~~~~~~~~
+
+- When designing an event, consider the consumers that will be using it. What information do they need to react to the event? What data is necessary for them to process the event?
```
sarina, line 117:
```suggestion
- When designing an event, consider the consumers that will be using it. What information do they need to react to the event? What data is necessary for them to process the event?
- You can't always predict the needs of consumers in the future. Ensure your design is discrete, flexible, and well-documented, so new and novel use cases can be developed.
```

sarina (approved):
Two comments - take them, leave them, or adjust them - but otherwise looks good to me.

docs/decisions/0016-event-design-practices.rst
mariajgrimaldi: Yes, it helps! Thank you. So I'll go ahead and create a new ADR considering the full transitive evolution strategy, so at least it is documented somewhere, noting that we haven't gone through any process of evolution just yet.

mariajgrimaldi:
This is a gentle reminder, @robrap, @Alec4r, and @bmtcril. Can you help review/approve this PR? Let me know if you agree with the ADR. Thanks!

Alec4r (approved):

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,127 @@
+16. Event Design Practices
```
robrap, line 1: Nit: => Event Design Best Practices? Were you intentionally avoiding calling these "best practices"?

docs/decisions/0016-event-design-practices.rst
```diff
@@ -0,0 +1,127 @@
...
+Where:
+
+- The event name indicates what happened: ``STUDENT_REGISTRATION_COMPLETED``.
+- The event description explains why the event happened: ``emitted when the user registration process in the LMS is completed``.
+- The event data contains data directly related to what happened ``UserData`` which should contain the necessary information to understand the event, like the username and email of the user.
+
+Responsibility and Granularity
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Design events with a single responsibility in mind. Each event should represent a single action or fact that happened in the system. If an event contains multiple actions, consider splitting it into multiple events. For instance, if the course grade is updated to pass or fail, there should be two events: one for the pass action and another for the fail action.
```
robrap, line 58: I know this ADR is about the event design itself, but somewhere you may want to consider adding a note like the following:

Events that are split across multiple actions is an exceptional case where the same event queue/topic should be used to help maintain order across these events.

robrap (approved):
Thanks @mariajgrimaldi. Looks good. I don't have any blocking comments. Just a variety of thoughts related to the docs in general.

This doc made me wonder about how-to vs adr when it comes to these docs. I don't think there is any perfect answer. I think the most important thing is helping people find this information when they need it.

Related:
1. We don't have adr vs how-to for OEPs, so there, we sort of use the OEP type of "Architecture" vs "Best Practice", and even there it gets fuzzy which one to use at times.
2. We used OEPs for cross-open-edx decisions for events before this repo existed. For example: https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0041-arch-async-server-event-messaging.html. I don't know if people would know that event related decisions from before a specific date are listed as OEPs, and after a certain date are dropped as ADRs in this repo. This is something you should consider as part of the FC. You could drop notes in the OEPs stating where all future event-related ADRs live, or you could migrate + note, etc.

sarina:
> I don't know if people would know that event related decisions from before a specific date are listed as OEPs, and after a certain date are dropped as ADRs in this repo. This is something you should consider as part of the FC. You could drop notes in the OEPs stating where all future event-related ADRs live, or you could migrate + note, etc.

I like the idea of adding a brief update to the OEP to direct people to look in this repo for ADRs.

docs/decisions/0016-event-design-practices.rst
mariajgrimaldi: This note's pretty useful! Thanks. I'll make sure to add it right away.

bmtcril (approved):

PR #505, opened 2025-06-09
chore: Add references to docs to make relative, not direct, references

sarina:
Usage of `:doc:` is an antipattern. It is fragile and prone to breaking cross references when docs are moved or renamed.

Adding in `.. _reference:` syntax to files and headings means cross-references can instead be made with the `:ref:` directive, which will (presuming the references themselves are not deleted or renamed) be more robust to docs refactorings.

### DIFF
--- a/docs/concepts/event-bus.rst
+++ b/docs/concepts/event-bus.rst
@@ -1,5 +1,7 @@
 .. include:: ../common_refs.rst
 
+.. _Event Bus:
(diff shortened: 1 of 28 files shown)


sarina:
@openedx/hooks-extension-framework this is ready for review! It's blocking a downstream PR so approval/merge would be appreciated.

PR #559, opened 2026-04-06
fix: restore all events in OpenEdxEventsTestMixin.tearDownClass

feanil:
## Summary

`OpenEdxEventsTestMixin.setUpClass()` disables all OpenEdX events then enables only those in `ENABLED_OPENEDX_EVENTS`, but had no `tearDownClass` to restore event state. This meant that after any test class using this mixin completed, all events not in `ENABLED_OPENEDX_EVENTS` remained globally disabled for the remainder of the process.

This was discovered in openedx-platform when shard rebalancing caused `lms/djangoapps/certificates/` and `lms/djangoapps/discussion/django_comment_client/` to run in the same process for the first time. Nine certificate test classes with `ENABLED_OPENEDX_EVENTS = []` left all events disabled, causing `ForumEventTestCase.test_comment_event` to fail with:

```
AttributeError: 'NoneType' object has no attribute 'kwargs'
```

because `FORUM_RESPONSE_COMMENT_CREATED` was never fired (it had been disabled).

## Fix

* Add `tearDownClass` to `OpenEdxEventsTestMixin` that calls `enable_all_events()`, mirroring the disable in 
`setUpClass`. This is consistent with the pattern already used by `FreezeSignalCacheMixin` in the same file.
* This also fixes a bug introduced in https://github.com/openedx/openedx-events/pull/557 which will break `openedx-platform` when it's released. See the refactor commit message for more details.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,16 @@ Unreleased
 __________
 
 
+[11.1.1] - 2026-04-06
+---------------------
+
+Fixed
+~~~~~
+
+* ``OpenEdxEventsTestMixin`` now re-enables all events in ``tearDownClass`` to
+  prevent disabled event state from leaking into subsequent test classes running
+  in the same process.
+
 [11.1.0] - 2026-03-19
 ---------------------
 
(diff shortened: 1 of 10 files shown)


mariajgrimaldi (approved):
Thanks a lot!

mariajgrimaldi:
https://github.com/openedx/openedx-events/releases/tag/v11.1.1 🥇

Issue #563, opened 2026-04-14
Add type annotations


bradenmacdonald:
Attempting to use this library in a project that uses mypy (e.g. `openedx-core`) results in a mypy error because it doesn't seem to have any type annotations:

<img width="1001" height="101" alt="Image" src="https://github.com/user-attachments/assets/e8c2543c-2717-483d-bce1-5367ec3b7c3d" />

```
signals.py:6: error: Skipping analyzing "openedx_events.tooling": module is installed, but missing library stubs or py.typed marker  [import-untyped]
signals.py:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 247 source files)
```

To fix this:
1. Add type annotations, and
2. Add a `py.typed` marker so the annotations will get installed and used.

bradenmacdonald:
There are many different ways this could be implemented. Here is just one example, that provides the best typing experience across the board and could be fully backwards-compatible, but requires transitioning the receivers to split up `data` and `metadata` as separate args instead of combining all the data+metadata fields into a single `kwargs` (the combined `kwargs` makes typing way more difficult, but could continue to be provided as well for some time to not break anything).

```python

_T = TypeVar("_T")

class OpenEdxPublicSignal(Signal, Generic[_T]):
    def send_event(self, *, data: _T, send_robust: bool = True, time=None) -> list: ...

@attrs.define(frozen=True)
class XBlockCreatedEventData:
    xblock_info: XBlockData

XBLOCK_CREATED: OpenEdxPublicSignal[XBlockCreatedEventData] = OpenEdxPublicSignal(
    event_type="org.openedx.content_authoring.xblock.created.v1",
    data_cls=XBlockCreatedEventData,  # replaces the dict
)
```

Sending:

```python
XBLOCK_CREATED.send_event(
    data=XBlockCreatedEventData(xblock_info=xblock_data),  # Fully typed
    time=now,
)
```

Receiving:

```
def on_xblock_created(
    sender,
    data: XBlockCreatedEventData,  # fully typed
    metadata: EventsMetadata,      # always the same, no per-signal variation
    **kwargs,
): ...
```

bmtcril:
I have an old PR for this that I can resurrect. I'd like to tackle it in 2 stages:

1- Add type hints without changing any logic / data structures except internal APIs to fix type problems
[...]

PR #565, opened 2026-04-17
Add type hinting across the package

bmtcril:
## Description

[This issue](https://github.com/openedx/openedx-events/issues/563) came up, and I had an old set of related changes. This PR should update the code to be mypy compliant and include typing for downstream packages.  Several other style issues were fixed along the way, so it is a large set of changes, but should not introduce any new behaviors. If downstream code was using the existing type hinting previously, it may fail linting on these changes and require updates.

Notes:
- There is a significant change to user `attrs` instead of the older `attr.s` for clearer semantics (simply use type hinting instead of explicitly declaring the type via  `attr.ib(...)` in most cases
- There is significant use of `Any` in this PR due to several places in the code that take wildly varied inputs. This can be tightened up, but I wanted to get this first pass up for review before spending any more time on it
- Signals cannot currently be annotated as it breaks edx-lint's static annotation cherker, [this PR](https://github.com/openedx/edx-lint/pull/535) will unblock that work (I have it stashed). Sadly this is probably the most valuable part for @bradenmacdonald 
- Commits are made in hopefully reasonable chunks with little-to-no overlap, so it may be easier to review commit-by-commit.

## Testing instructions

I'll be adding PRs to the event bus repos and platform to see what the downstream impact is shortly, this will stay in draft until then.

**Merge Checklist:**
- [ ] All reviewers approved
- [ ] Reviewer tested the code following the testing instructions
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added with short description of the change and current date
- [ ] Integration with other services reviewed
- [ ] Fixup commits are squashed away
- [ ] Unit tests added/updated
- [ ] Noted any: Concerns, dependencies, migration issues, deadlines, tickets

**Post Merge:**
- [ ] Create a tag
- [ ] Create a release on GitHub
      finished.
- [ ] Upgrade the package in the Open edX platform requirements (if applicable)

Claude Sonnet 4.6 was used to modernize changes from an old PR and break them up into smaller commits.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -52,10 +52,10 @@ disallow_untyped_defs = True
 disallow_untyped_defs = True
 
 [mypy-openedx_events.analytics.data]
-disallow_untyped_defs = False
+disallow_untyped_defs = True
 
 [mypy-openedx_events.enterprise.data]
-disallow_untyped_defs = False
+disallow_untyped_defs = True
 
 [mypy-openedx_events.content_authoring.data]
 disallow_untyped_defs = False
--- a/mypy.ini
+++ b/mypy.ini
@@ -61,7 +61,7 @@ disallow_untyped_defs = True
 disallow_untyped_defs = True
 
 [mypy-openedx_events.learning.data]
-disallow_untyped_defs = False
+disallow_untyped_defs = True
 
 [mypy-openedx_events.event_bus.avro.types]
 disallow_untyped_defs = False
--- a/mypy.ini
+++ b/mypy.ini
@@ -64,19 +64,19 @@ disallow_untyped_defs = True
 disallow_untyped_defs = True
 
 [mypy-openedx_events.event_bus.avro.types]
-disallow_untyped_defs = False
+disallow_untyped_defs = True
 
 [mypy-openedx_events.event_bus.avro.custom_serializers]
-disallow_untyped_defs = False
+disallow_untyped_defs = True
 
 [mypy-openedx_events.event_bus.avro.schema]
-disallow_untyped_defs = False
+disallow_untyped_defs = True
 
 [mypy-openedx_events.event_bus.avro.serializer]
-disallow_untyped_defs = False
+disallow_untyped_defs = True
 
 [mypy-openedx_events.event_bus.avro.deserializer]
-disallow_untyped_defs = False
+disallow_untyped_defs = True
 
 [mypy-openedx_events.tooling]
 disallow_untyped_defs = False
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,109 @@
+[mypy]
+python_version = 3.12
+; Start permissive at the project level — tighten per-module as each phase lands
+warn_return_any = True
+warn_unused_ignores = True
+warn_unused_configs = True
+warn_redundant_casts = True
+disallow_untyped_defs = False
+disallow_incomplete_defs = False
+check_untyped_defs = True
+no_implicit_optional = True
+strict_equality = True
+
+plugins =
+    mypy_django_plugin.main
+
+[mypy.plugins.django-stubs]
+django_settings_module = test_utils.test_settings
+
+; -----------------------------------------------------------------------
+; Third-party packages without type stubs — silence missing-import errors
+; -----------------------------------------------------------------------
+
+[mypy-fastavro.*]
+ignore_missing_imports = True
+
+[mypy-ccx_keys.*]
+ignore_missing_imports = True
+
+[mypy-opaque_keys.*]
+ignore_missing_imports = True
+
+[mypy-edx_django_utils.*]
+ignore_missing_imports = True
+
+[mypy-ddt]
+ignore_missing_imports = True
+
+
+; -----------------------------------------------------------------------
+; Per-module strictness overrides
+; Flip disallow_untyped_defs = True module-by-module as annotations land.
+; -----------------------------------------------------------------------
+
+[mypy-openedx_events.exceptions]
+disallow_untyped_defs = False
+
+[mypy-openedx_events.utils]
+disallow_untyped_defs = False
+
+[mypy-openedx_events.data]
+disallow_untyped_defs = False
+
+[mypy-openedx_events.analytics.data]
+disallow_untyped_defs = False
+
+[mypy-openedx_events.enterprise.data]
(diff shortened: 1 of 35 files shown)


bmtcril:
Kafka test PR is passing (though it was way behind and broken so that PR has a number of other fixes): https://github.com/openedx/event-bus-kafka/pull/344

bmtcril:
Redis bus PR: https://github.com/openedx/event-bus-redis/pull/228

mypy.ini
```diff
@@ -0,0 +1,109 @@
...
+; -----------------------------------------------------------------------
+
+[mypy-fastavro.*]
+ignore_missing_imports = True
+
+[mypy-ccx_keys.*]
+ignore_missing_imports = True
+
+[mypy-opaque_keys.*]
+ignore_missing_imports = True
```
bradenmacdonald: opaque-keys has type info now, so we shouldn't need this.

openedx_events/data.py
```diff
@@ -71,62 +124,63 @@ class EventsMetadata:
...
 
-    event_type = attr.ib(type=str, validator=attr.validators.instance_of(str))
-    id = attr.ib(
-        type=UUID, default=None,
-        converter=attr.converters.default_if_none(attr.Factory(lambda: uuid1())),  # pylint: disable=unnecessary-lambda
-        validator=attr.validators.instance_of(UUID),
+    event_type: str = attrs.field(validator=attrs.validators.instance_of(str))
+    id: UUID = attrs.field(
+        default=None,
+        converter=_default_id,
```
bradenmacdonald, line 130: Why this `converter` pattern instead of


```suggestion
        factory=uuid1,
```

or 

```suggestion
        default=attrs.Factory(uuid1),
```
? I think using `default` or `factory` like this is simpler, and you don't need to implement the `_default_x` functions yourself.

bradenmacdonald (commented):
Wow, awesome work - thanks! ✅ LGTM but I'm not a committer on this repo :)

bmtcril:
@openedx/committers-openedx-events can folks offer opinions on whether this is a good direction? I can rebase and wrap it up if so.

openedx_events/event_bus/avro/custom_serializers.py
```diff
@@ -46,12 +47,12 @@ class CourseKeyAvroSerializer(BaseCustomTypeAvroSerializer):
     field_type = PYTHON_TYPE_TO_AVRO_MAPPING[str]
 
     @staticmethod
-    def serialize(obj) -> str:
+    def serialize(obj: Any) -> str:
```
navinkarkera, line 50:
```suggestion
    def serialize(obj: CourseKey) -> str:
```

It this doesn't fail mypy check, we can do the same for all classes below.

openedx_events/content_authoring/data.py
```diff
@@ -103,10 +104,10 @@ class DuplicatedXBlockData(XBlockData):
         source_usage_key (UsageKey): identifier of the source XBlock object.
     """
 
-    source_usage_key = attr.ib(type=UsageKey)
+    source_usage_key: UsageKey = attrs.field(kw_only=True)
```
navinkarkera, line 107: Is change to `kw_only=True` intentional?

openedx_events/learning/data.py
```diff
@@ -571,10 +571,10 @@ class CcxCoursePassingStatusData(CoursePassingStatusData):
         All other attributes are inherited from CoursePassingStatusData.
     """
 
-    course = attr.ib(type=CcxCourseData)
+    course: CcxCourseData = attrs.field(kw_only=True)  # type: ignore[assignment]
```
navinkarkera, line 574: Same here, is this intentional?

navinkarkera (approved):
@bmtcril Looks good. Thanks!

openedx_events/data.py
bmtcril: The default + converter lets us handle `None` better, where the factory pattern would ignore any place where we explicitly pass `None` (`tooling.py:88`) vs just omitting the param. The converter runs before the validator so it produces an actual value. I wasn't sure if we have other places where we might pass in explicit `None`  but this seemed safer.

openedx_events/event_bus/avro/custom_serializers.py
bmtcril: This would be somewhat weird due to overriding the type of the base class, and since callers don't access them directly it wouldn't help us much.

openedx_events/content_authoring/data.py
bmtcril: Yes, these end up being needed due to subclasses having fields that are required for them, but the base class has optional fields. So without it the field becomes positional but after the optional `value` on the base class and a ValueError gets thrown that "no mandatory attributes allowed after an attribute with a default".

openedx_events/learning/data.py
bmtcril: It is, but the reason is different. When you re-declare an attribute attrs removes it and appends it to the end of the positional list. So if we didn't do this the order of args for `CoursePassingStatusData` would be `(is_passing, course, user)` but `CcxCoursePassingStatusData` would be `(is_passing, user, course)`. So making it kw_only takes the ordering out of the equation to prevent that specific issue. 

It does mean checking the existing calls to those, though, which I'll do shortly.

bmtcril:
This is the openedx-platform PR that tests these changes. There were a very few places that needed updating:

https://github.com/openedx/openedx-platform/pull/38892

bmtcril:
@mariajgrimaldi @felipemontoya - last check here, I've got a thumb but wanted to make sure you have a chance to take a look before I merge it.

@robrap - this is just a heads up in case you wanted to look for potential bus issues

robrap:
> @robrap - this is just a heads up in case you wanted to look for potential bus issues

@bmtcril: Thanks for the heads up. I don't plan on reviewing in detail, and hope the tests were enough to keep things backward-compatible. Good luck.

felipemontoya:
Thanks @bmtcril, this looks like a good improvement to me. I only looked at it statically in the PR, I did not launch a testing environment.

Given that it will break linting for downstream users I assume we would bump the major version when releasing. Is that correct?

Ian2012 (approved):
LGTM, nice change

